PR #6, opened 2016-04-09
Operations

ravisantoshgudimetla:
Created a initial operations with following features.

-Provision
-Delete
-Remove
-Snapshot
-Clone


docs/operations.py
```diff
@@ -0,0 +1,180 @@
...
+def provision(node_name, img_name = "hadoopMaster.img",\
+        snap_name = "HadoopMasterGoldenImage",\
+        debug = False):
+    try:
+        fsconfig = create_fsconfigobj()
+        fs_obj = init_fs(fsconfig)
+        if fs_obj.clone(img_name.encode('utf-8'),\
+                snap_name.encode('utf-8'),\
+                node_name.encode("utf-8")):
+            iscsi_output = call_shellscript('/home/user/ims/iscsi_update.sh', \
```
genagain: How do we know that the `iscsi_update.sh` file will always be in the `/home/user/ims` directory? Is it possible that the `iscsi_update.sh` file could be in a different user's directory?

ravisantoshgudimetla: @genagain Valid point, thats something that we need to fix but this is more less going to be a VM that will be downloaded by user and this has prepackaged stuff, so the path may not change at all but we will fix that out.

suranapranay:
@genagain , correct!. I do not know why we are not using configuration files. We actually have those.
The Question you should ask is, "WHY? are we using a shellscript in the first place! It is horrible.

ravisantoshgudimetla:
@suranapranay : I think you are talking about bmiconfig.cfg file. We are reading all these properties through python. Correct me if I am wrong. This shell script(iscsi_update.sh) is for rbd map which we don't have in python wrapper for librados.

suranapranay:
@ravisantoshgudimetla . I know we discussed this, I'm just putting this here for future reference. The shell script apparently also updates some config file, this we should do using python.

suranapranay:
@ravisantoshgudimetla , yes I'm talking about the bmiconfig. we can use that for all paths etc. Each of our function uses it already.

docs/operations.py
```diff
@@ -0,0 +1,180 @@
...
+           type(rbd.ImageExists())  : 401,
+           type(CephException()) : 403,
+           type(rbd.ImageBusy()) : 409,
+           type(rbd.ImageHasSnapshots()) : 405,
+           type(rbd.ImageNotFound()) : 404,
+           type(rbd.FunctionNotSupported()) : 410,
+           type(rbd.ArgumentOutOfRange()) : 411,
+           type(ShellScriptException()) : 440,
+                                }
+        #This is to handle key error exception, this also gives us the default error
```
suranapranay: I strongly disagree with this. This should not be done. Each and every exception should exist in the dictionary.

ravisantoshgudimetla: I think there is a bit a confusion here. The exception that we are seeing here are user specific, whatever the end-user sees. Imagine there are thousands of exceptions that are used by python like Filenotfound, keyerror etc. Are you telling to handle all these things and let the end-user know all these exceptions. I guess this is something that we have agreed upon and decided to use it the existing way...

ravisantoshgudimetla:
@suranapranay : Coming to shell script. I think we discussed that as well. The iscsi config file that it updates is coming from device name that is mapped. So, in a way if we use python, we should create some other shell script which greps for nodename in mapped devices and gives the value, so that this can be updated in iscsi config file.- Thats the reason we are updating using shell.

docs/operations.py
suranapranay: There is no confusion. Python should not be producing any other exception, whatsoever. Doing this is akin to doing a try"whole program" catch "do nothing, return 500". This is wrong. The end use will never see any of the other exceptions, because they will never be generated. I strongly disagreed then. I disagree now too.

ravisantoshgudimetla:
Guys, I moved the file one directory up, so that docs will have only documentation. Its a mistake from my side. Request to check on latest code for operations.

genagain:
Looks good to me! +1

anuj-rt:
Looks good to me! +1

anuj-rt:
The changes look good to me! +1

PR #7, opened 2016-04-09
Ceph driver

ravisantoshgudimetla:
This file is ceph wrapper around librados and librbd API that we created for interactions with ceph.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,60 @@
+# BMI
+
+The Bare Metal Imaging (BMI) is a core component of the Massachusetts Open Cloud and a image management system(ims) that 
+(i)provisions numerous nodes as quickly as possible while preserving support for multitenancy using Hardware as a Service (HaaS) and 
+(ii)introduces the image management techniques that are supported by virtual machines, with little to no impact on application performance.
+
+Motivation
+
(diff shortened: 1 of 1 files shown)


genagain:
Looks good to me! +1

anuj-rt:
#7 
remove() method in RBD class(ceph_wrapper.py) will print extra ' (single inverted comma) in the Ceph exception messages.

anuj-rt:
Rest of the stuff looks good. +1

ceph_wrapper.py
```diff
@@ -0,0 +1,217 @@
...
+            if img_name not in self.img_dict:
+                raise CephException("Invalid image name"\
+                        " or image not instantiated")
+            snap_list_iter = self.img_dict[img_name].list_snaps()
+            snap_list = [snap['name'] for snap in snap_list_iter]
+            return snap_list
+        except Exception as e:
+            raise e
+    
+    def remove_snapshots(self, img_name, name, debug = False):
```
genagain: Maybe I'm missing something, but it doesn't seem like this method can remove more than one snapshot at a time. Because of this could the "s" in the method name be removed, so it's `remove_snapshot`? I know this is pretty minor, but it may be confusing for someone else using this method.

genagain:
The two new methods look good to me! +1

anuj-rt:
The functionality looks good to me! +1

PR #9, opened 2016-04-21
Only documentation changes

ravisantoshgudimetla:
Documentation for only changes related to API

### DIFF
--- a/docs/rest_api.md
+++ b/docs/rest_api.md
@@ -1,0 +1,253 @@
+We have a basic set of API as of now. All of the services as of now are taking the arguments as parameters and return response.
+
+BMI API assumes that you know the image which you are interested in. BMI has an driver which can communicate with Ceph(the file storage system that we are using for storing our images). Please make sure that you understand Ceph concepts before you play with the API. 
+
+To have a clear understanding of the API, we wish to provide the following terminology:
+
+project - A project is the HaaS project that has been allocated to tenant.
+node - A node that is allocated to project by HaaS. This node should be present in HaaS.
+img - The name of image which will be used for provisioning.
+snap_name - The snapshot of the image from which you want to provision a node.
+
+The convention that we are following for requests is:
+
+PUT - for resource creation like node creation 
+DELETE - for resource deletion for node deletion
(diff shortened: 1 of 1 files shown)


docs/rest_api.md
```diff
@@ -0,0 +1,253 @@
...
+http://BMI_SERVER:PORT/remove_snapshot
+with DELETE request type and request body as:
+
+body:
+
+{
+ "img" : "<img_name>" ,
+ "project" : "<project_name>" ,
+ "snap_name" : "snapshot_name>"
+}
```
gsilvis: DELETE calls are not allowed to have bodies

docs/rest_api.md
```diff
@@ -0,0 +1,253 @@
...
+"snap_name":"test_snap2016" 
+}
+
+This should return a 200 or other errors as explained above.
+
+List snapshots:
+
+We can have as many snapshots as possible for a particular image. So, using this we can list all the snapshots available for an image.
+
+http://BMI_SERVER:PORT/list_snapshots
```
gsilvis: This call does not change any state on the server, so it should be a `GET`

gsilvis:
This API is not using the various REST calls correctly.  Here are the basic rules you should follow:

- For any informational operation that does not change the state of the server, use `GET`
- To create a resource, use `PUT` on a path that will represent the resource: for example, `PUT /images/<new_image_name>`
- Another way to create a new resource is to use a `POST` on the directory that will contain the new resource:  `POST /images`
- To delete a resource, use `DELETE` on the path that represents the resource: `DELETE /images/<image_name>`
- For truly miscellaneous operations (e.g. 'power_cycle_node' in HaaS), use a POST call on a URL somehow linked to any applicable resource; e.g. `POST /images/<image_name>/frobnicate_image`

gsilvis:
But, if the API itself is not up for debate in this PR, then this documentation looks okay

rahulait:
Agree with George. Ideally it should follow the REST semantics than the URLs. Maybe we can fix it now rather than doing in future? Rest looks ok to me.

rahulait:
+1

PR #10, opened 2016-04-22
Api

ravisantoshgudimetla:
API default.py file for webservices


chemistry-sourabh:
@ravisantoshgudimetla @apoorvemohan  updated PR

docs/rest_api.md
```diff
@@ -0,0 +1,289 @@
...
+ "img" : "<image_name>" ,
+ "network" : "<network_name>" ,
+ "channel" : "<channel in haas>" ,
+ "nic" : "<nic to connect on>"
+}
+```
+
+####Responses:
+* 200. This means the provision call is successful.
+* Internal 500 with some junk characters. This means the request body is not proper.
```
knikolla: 500 means Internal Server Error. It means the request was proper but the server failed to process it for some fault of the server, not the client or the request.

The correct return code here is [400](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error).

docs/rest_api.md
```diff
@@ -0,0 +1,289 @@
...
+* 401. This means unauthorized access to ceph image or snapshot or image already exists in ceph.
+* 403. This means a ceph connection problem.
+* 409. Image busy exception.
+* 444. You used a wrong request method like PUT instead of POST etc.
```
knikolla: [405 - Method not allowed](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes#4xx_Client_Error).

docs/rest_api.md
```diff
@@ -0,0 +1,289 @@
...
+###Provision:
+Provision API is needed for provisioning a node from MOC cluster as of now. Provision operation internally calls a ceph clone operation. Ceph clone operation usually takes a snap_
+
+Following is the call for API:
+
+####Link:
+http://BMI_SERVER:PORT/provision/
+
+####Request Type:
+PUT
```
knikolla: Technically, the correct method in these cases should have been POST. See [here](http://restcookbook.com/HTTP%20Methods/put-vs-post/).

docs/rest_api.md
```diff
@@ -0,0 +1,289 @@
...
+ "network" : "<network_name>" ,
+ "channel" : "<channel in haas>" ,
+ "nic" : "<nic to connect on>"
+}
+```
+
+####Responses:
+* 200. This means the provision call is successful.
+* Internal 500 with some junk characters. This means the request body is not proper.
+* 401. This means unauthorized access to ceph image or snapshot or image already exists in ceph.
```
knikolla: This should be 403 Forbidden.

docs/rest_api.md
```diff
@@ -0,0 +1,289 @@
...
+* 403. This means a ceph connection problem.
```
knikolla: This is clearly a server error, so 500.

naved001:
I am closing this because we merged Sourabh's TGT branch to the dev branch #20

PR #12, opened 2016-05-13
Database

chemistry-sourabh:


database/repositories/image_repository.py
```diff
@@ -0,0 +1,87 @@
+from database.tables.image import *
+
+
+# This class is responsible for doing CRUD operations on the Image Table in DB
+# This class was written as per the Repository Model which allows us to change the DB in the future without changing
+# business code
+class ImageRepository:
+    def __init__(self):
+        self.connection = None
```
knikolla: Why do you have this here if on every call you create a new one?

database/repositories/image_repository.py
```diff
@@ -0,0 +1,87 @@
...
+            img = Image()
+            img.name = image_name
+            img.project_id = project_id
+            img.is_public = is_public
+            self.connection.session.add(img)
+            self.connection.session.commit()
+        # should change to more specific exception
+        except Exception:
+            self.connection.session.rollback()
+            raise
```
knikolla: Maybe raise a more specific Exception?

database/repositories/image_repository.py
```diff
@@ -0,0 +1,87 @@
...
+        finally:
+            self.connection.close()
+
+    # deletes images with name under the given project name
+    # commits if deletion was successful otherwise rollback occurs and exception is bubbled up
+    def delete_with_name_from_project(self, name, project_name):
```
knikolla: I feel like all those methods should have a decorator, and handle the exception catching and rollback there.

database/repositories/image_repository.py
```diff
@@ -0,0 +1,87 @@
...
+    # returns a array of image ids of the images which have the given name
+    def fetch_id_with_name_from_project(self, name, project_name):
+        try:
+            self.connection = DatabaseConnection()
+            for image in self.connection.session.query(Image).filter_by(name=name):
+                if image.project.name == project_name:
+                    return image.id
+        # should change to more specific exception
+        except Exception:
+            print "Database Exception: Something bad happened related to database"
```
knikolla: Whom are you printing to? Is this a console application?

database/repositories/image_repository.py
```diff
@@ -0,0 +1,87 @@
...
+
+    # Fetch the list of images which are public
+    # We are returning a dictionary of format {image_name : <img_name> , project_name : <proj_name>}
+    def fetch_name_with_public(self):
+        try:
+            img_list = []
+            self.connection = DatabaseConnection()
+            for image in self.connection.session.query(Image).filter_by(is_public=True):
+                img_list.append(image)
+                return [{'image_name': img.name, 'project_name': img.project.name} for img in img_list]
```
knikolla: This is wrong and will return only the first image because you are returning inside the loop.

Please change the entire for loop to:
```
img_list = self.connection.session.query(Image).filter_by(is_public=True)
return [{'image_name': img.name, 'project_name': img.project.name} for img in img_list]
```

database/test_database.py
```diff
@@ -0,0 +1,103 @@
...
+        # check whether second project is present
+        pid = pr.fetch_id_with_name("project 2")
+        self.assertIsNotNone(pid)
+        self.assertEqual(pid, 2)
+
+        # verify that None is being returned for a project which doesnt exist
+        pid = pr.fetch_id_with_name("project 3")
+        self.assertIsNone(pid)
+
+    # should not actually be called
```
knikolla: not necessary to mention, it's clear it's a test.

database/test_database.py
```diff
@@ -0,0 +1,103 @@
...
+
+        # verify that it is gone
+        pid = pr.fetch_id_with_name("project 1")
+        self.assertIsNone(pid)
+
+    # should not be called
+    # runs the project test case
+    def project_test(self):
+        self.insert_project_test()
+        self.delete_with_name_project_test()
```
knikolla: You don't need to call the test methods. The test framework will automatically call all methods that start with test. So please rename all the test methods to start with test. 

https://docs.python.org/2.7/library/unittest.html

database/test_database.py
```diff
@@ -0,0 +1,103 @@
...
+        self.assertEqual(pid, 2)
+
+        # verify that None is being returned for a project which doesnt exist
+        pid = pr.fetch_id_with_name("project 3")
+        self.assertIsNone(pid)
+
+    # should not actually be called
+    # the test for delete
+    # delete the inserted row
+    # check if it is gone
```
knikolla: The comments don't really add any value. Please only use comments when you feel it is necessary to explain something not clear in code.

database/test_database.py
```diff
@@ -0,0 +1,103 @@
+from unittest import TestCase
+
+from database.image import *
+from database.project import *
+
+
+# Before running make sure no .db files are present in execution directory
```
knikolla: Please, to run the unit tests use an in memory sqlite database. https://www.sqlite.org/inmemorydb.html

chemistry-sourabh:
@knikolla updated the branch please have a look

knikolla:
+1, looks good enough for now.

ravisantoshgudimetla:
I saw that too @knikolla. In our case we are explicitly closing DB connection as well. 
But it seems we can use in memory DB.  I ran the test cases with something called staticpool. You can find how to use it here.

http://www.sameratiani.com/2013/09/17/flask-unittests-with-in-memory-sqlite.html

knikolla:
@ravisantoshgudimetla awesome! Next thing you want to look at is have travis-ci automatically run the unit tests for each pull request. I can help with that if you want.

ravisantoshgudimetla:
@knikolla : Thanks a lot for reviewing the code and helping us standardize code. We are working on jenkins integration so that we can run some tests against ceph. We have something in place. Need to test out fully. Will let you know in case we need help. Thanks for offering help again.

gsilvis:
It looks like @knikolla has already made almost all the comments I'd make.  My only request is:

When you have a query that should always return either 0 or 1 results, you should use `.one()` rather than `.first()`.  Just like `.first()`, it raises an exception if there are none; but it also raises an exception if there is more than one.  (I'm thinking especially of when you're looking up an object by name, and you have a uniqueness constraint on names.)

chemistry-sourabh:
Done have a look @gsilvis

ims/database/image.py
```diff
@@ -46,12 +46,10 @@ def delete_with_name_from_project(self, name, project_name):
...
-            images = self.connection.session.query(Image). \
-                filter(Image.project.has(name=project_name)).filter_by(name=name)
-            if images.count() == 0:
-                return None
-            else:
-                return images.first().id
+            image = self.connection.session.query(Image). \
+                filter(Image.project.has(name=project_name)).filter_by(name=name).one_or_none()
+            if image is not None:
+                return image.id
```
knikolla: There is no need to check if the image is None. Just return it. You are returning nothing anyway in that case.

ims/database/project.py
```diff
@@ -49,11 +47,9 @@ def delete_with_name(self, name):
...
         try:
             self.connection = DatabaseConnection()
-            projects = self.connection.session.query(Project).filter_by(name=name)
-            if projects.count() == 0:
-                return None
-            else:
-                return projects.first().id
+            project = self.connection.session.query(Project).filter_by(name=name).one_or_none()
+            if project is not None:
+                return project.id
```
knikolla: Same comment as above.

gsilvis:
Looks good to me.  +1

ims/database/image.py
gsilvis: `image.id` won't work if `image` is `None`

ims/database/project.py
gsilvis: ditto

chemistry-sourabh:
Thanks a lot for the feedback @knikolla and @gsilvis

PR #13, opened 2016-05-28
Exceptions and Beautification

chemistry-sourabh:


bmiconfig.cfg
```diff
@@ -0,0 +1,20 @@
...
+
+# This section is for haas related config
+[haas]
+url = <base url for haas>
+
+# This section is for iscsi related config
+[iscsi]
+# will be replaced with python script
+update_shell_url = <location of iscsci_update shell script>
+password = <sudo password for iscsi_update script>
```
gsilvis: All files should end in newlines

ims/exception/parser.py
```diff
@@ -0,0 +1,23 @@
...
+                    file_system_exceptions.ArgumentsOutOfRangeException: 477,
+                    db_exceptions.ImageNotFoundException: 404,
+                    db_exceptions.ProjectNotFoundException: 404,
+                    haas_exceptions.AuthenticationFailedException: 401,
+                    haas_exceptions.AuthorizationFailedException: 403}
+
+    # returns the status code as per the class of the exception
+    # default status code is 500
+    def parse(self, ex):
+        return ExceptionParser.status_codes.get(type(ex), 500)
```
gsilvis: Why not just have an attribute of your exception class that says what the status code is?  e.g. add a pure virtual to BMIException 'status_code' that child classes must implement.

gsilvis:
The `with` statements are nice. I like what you're doing with exceptions---just one comment there.  The changes to actual BMI logic are too involved for me to figure out what's going on---several files were entirely re-written, and I wasn't familiar with them beforehand.

chemistry-sourabh:
@gsilvis  changed the exceptions as per your suggestion. Please have a look when you have time.

gsilvis:
Exceptions look good

ims/database/image.py
```diff
@@ -7,77 +8,81 @@
...
-        finally:
-            self.connection.close()
+    def insert(self, image_name, project_id, is_public=False, id=None):
+        with DatabaseConnection() as connection:
+            try:
+                img = Image()
+                img.name = image_name
+                img.project_id = project_id
+                img.is_public = is_public
+                if id is not None:
```
kamfonik: What is the use case for specifying 'id' here - isn't id your image table's primary key?  Same

ims/exception/file_system_exceptions.py
```diff
@@ -0,0 +1,129 @@
...
+
+
+# This exception should be raised when Image is not found in file system
+class ImageNotFoundException(FileSystemException):
+    @property
+    def status_code(self):
+        return 404
+
+    def __init__(self, name):
+        self.name = name
```
kamfonik: Very minor style/readability note, but it seems what you actually pass to this is actually an image ID, not an image name.

ims/database/image.py
chemistry-sourabh: You are right it is our primary key. By default the id is auto incremented , we added it in the method so that we can use it in test cases as the id is used for mapping the image name in db to the image in ceph.

ims/exception/file_system_exceptions.py
chemistry-sourabh: The image id in db is the image name in ceph, so it is fine.

ims/database/image.py
kamfonik: Yes, that explains why you might specify it when testing.

kamfonik:
+1 This looks good to me, as far as I could follow without being familiar with the codebase. Sourabh addressed my comments above.

chemistry-sourabh:
thanks @kamfonik

apoorvemohan:
+1. My comments were addressed.

PR #18, opened 2016-07-27
Logging database

chemistry-sourabh:


ims/database/db_connection.py
```diff
@@ -3,6 +3,10 @@
 from sqlalchemy.orm import sessionmaker
 from sqlalchemy.pool import NullPool
 
+import ims.common.config as config
+
+_cfg = config.get()
```
ravisantoshgudimetla: Why are we using a global variable here? Can this be avoided?

ims/database/db_connection.py
```diff
@@ -15,18 +19,15 @@ class DatabaseConnection:
     # sample_bmi.db should be changed to something more realistic
     # NullPool pool class is equivalent to no connection pool
     # Should be adapted to postgres SQL
-    engine = create_engine('sqlite:///sample_bmi.db', poolclass=NullPool)
+    engine = create_engine('sqlite:///' + _cfg.db_url, poolclass=NullPool)
```
ravisantoshgudimetla: Will this be sqlite again, even if we change it to postgres? Shouldn't we change it to something generic in properties file to dynamically create a URL based on backend chosen?

ims/database/test_database.py
```diff
@@ -1,80 +0,0 @@
...
-        qimg = imgr.fetch_id_with_name_from_project("image 1", "project 2")
-        self.assertIsNone(qimg)
-
-        # testing on delete cascade
-        # Also brings the db to intial state
-        pr = ProjectRepository()
-        pr.delete_with_name("project 2")
-
-        qp = pr.fetch_id_with_name("project 2")
-        self.assertIsNone(qp)
```
ravisantoshgudimetla: Any specific reason for removing the test cases file?

naved001:
I am closing this because we merged Sourabh's TGT branch to the dev branch #20

PR #24, opened 2016-12-29
ISCSI Driver

chemistry-sourabh:
Ignore Changes to Tests
See Changes to config.py
Fixed * imports
ISCSI obj is loaded dynamically based on config
Check Mock Driver that uses a sqlite db to persist targets
Changed TGT to sh (the main ones) still need to complete transformation
Ignore IET Driver it is broken for now


chemistry-sourabh:
@ravisantoshgudimetla  check this out

ravisantoshgudimetla:
@chemistry-sourabh , Can you split this PR, this seems overwhelming. I would recommend to have something like a PR for each change that has gone in instead of single PR that changes everything.

ims/cli/cli.py
```diff
@@ -2,33 +2,36 @@
...
-import ims.common.config as config
 import ims.common.constants as constants
+import ims.common.config as config
+
 
 config.load()
 
 from ims.einstein.operations import BMI
-from ims.database import *
+# from ims.database.database import Database
```
ravisantoshgudimetla: Remove commented lines, easier to merge into master branch later.

ims/cli/cli.py
```diff
@@ -2,33 +2,36 @@
...
 
 _cfg = config.get()
 
-_url = "http://{0}:{1}/".format(_cfg.bind_ip, _cfg.bind_port)
+_url = "http://{0}:{1}/".format(_cfg.http[constants.BIND_IP_KEY],
+                                int(_cfg.http[constants.BIND_PORT_KEY]))
 
-if constants.HAAS_USERNAME_VARIABLE in os.environ:
-    _username = os.environ[constants.HAAS_USERNAME_VARIABLE]
+if constants.HIL_USERNAME_VARIABLE in os.environ:
```
ravisantoshgudimetla: I think we should agree upon some naming convention for variables, probably no need to have a "_VARIABLE" at last as this seems to be loaded from constants file.

ims/common/config.py
```diff
@@ -32,82 +33,50 @@ class __BMIConfig:
...
-            self.ipxe_loc = config.get(constants.TFTP_CONFIG_SECTION_NAME,
-                                       constants.IPXE_URL_KEY)
-
-            for k, v in config.items(constants.FILESYSTEM_CONFIG_SECTION_NAME):
-                if v == 'True':
-                    self.fs[k] = {}
-
-                    for key, value in config.items(k):
-                        self.fs[k][key] = value
         # Didn't Test
```
ravisantoshgudimetla: If you are concerned about how to test this, the ideal way would be to load a config file that doesn't contain some of mandatory sections.

ims/common/config.py
```diff
@@ -32,82 +33,50 @@ class __BMIConfig:
...
-        self.pxelinux_loc = None
-        self.ipxe_loc = None
-
+        self.db = {}
+        self.logs = {}
+        self.rpc = {}
+        self.http = {}
+        self.iscsi = {}
+        self.tftp = {}
+        self.netiso = {}
```
ravisantoshgudimetla: What is netiso? Any reason for choosing this name?

ims/common/config.py
```diff
@@ -32,82 +33,50 @@ class __BMIConfig:
...
+            self.netiso = dict(
+                config.items(constants.NETISO_CONFIG_SECTION_NAME))
+            self.iscsi = dict(config.items(constants.ISCSI_CONFIG_SECTION_NAME))
+            self.tftp = dict(config.items(constants.TFTP_CONFIG_SECTION_NAME))
+            self.bmi = dict(config.items(constants.BMI_CONFIG_SECTION_NAME))
+            self.logs = dict(config.items(constants.LOGS_CONFIG_SECTION_NAME))
+            self.http = dict(config.items(constants.HTTP_CONFIG_SECTION_NAME))
+            self.rpc = dict(config.items(constants.RPC_CONFIG_SECTION_NAME))
+            self.db = dict(config.items(constants.DB_CONFIG_SECTION_NAME))
+
```
ravisantoshgudimetla: I like this way of dynamically loading stuff but 
1.) What about validations for keys defined in that dictionary, is there a way to do that now? Say something like iscsi_url is missing, how do we validate that?
2.) We are going to give a sample config file which contains sections, is there a need to define them in constants file again. Can we have single source of truth. Like a section name that exists in config file and get all the section dictionary after reading that file.

ims/common/constants.py
```diff
@@ -102,18 +99,18 @@
...
 
-HAAS_CALL_TIMEOUT = 10
+HIL_CALL_TIMEOUT = 10
 DEFAULT_SNAPSHOT_NAME = "snapshot"
 
 BMI_ADMIN_PROJECT = "bmi_infra"
 
-HAAS_BMI_CHANNEL = "vlan/native"
\ No newline at end of file
+HIL_BMI_CHANNEL = "vlan/native"
```
ravisantoshgudimetla: New line here

ims/database/db_connection.py
```diff
@@ -1,14 +1,22 @@
...
 
 
+# ISSUE
+# This Class is not creating the engine once as previously expected
+# Although there is no harm in creating the engine and session maker multiple
+# times, it doesnt make sense when it comes to multi threading stuff as there
+# is supposed to be one engine and one session per request.
+
+# Will Probably need to build Engine Abstraction To Solve it properly
+
```
ravisantoshgudimetla: Any specific reason for moving to this engine creation. I might have forgotten the discussion, so let me know, if this was something that was agreed upon.

ims/database/db_connection.py
```diff
@@ -20,7 +28,7 @@ class DatabaseConnection:
     # sample_bmi.db should be changed to something more realistic
     # NullPool pool class is equivalent to no connection pool
     # Should be adapted to postgres SQL
-    engine = create_engine('sqlite:///' + _cfg.db_url,
+    engine = create_engine('sqlite:///' + _cfg.db[constants.DB_URL_KEY],
```
ravisantoshgudimetla: I may have missed something here but what will happen, if constants.DB_URL_KEY doesn't exist? Do, we have some validation? (This is same as 1st validation that I told in previous  comments.)

ims/einstein/ceph.py
```diff
@@ -129,6 +129,7 @@ def remove(self, img_id):
         except rbd.ImageHasSnapshots:
             raise file_system_exceptions.ImageHasSnapshotException(img_id)
 
+    # Not Using Anywhere
```
ravisantoshgudimetla: These methods will be used when we have upload image ready.

ims/einstein/ceph.py
```diff
@@ -230,34 +231,46 @@ def get_parent_info(self, img_id):
...
-        if p.returncode == 0:
-            if output.find("sudo") != -1:
-                return output.split(":")[1].strip()
-            else:
+        try:
+            with sh.sudo:
+                output = sh.rbd.map(ceph_img_name, keyring=self.keyring,
+                                    id=self.rid,
+                                    pool=self.pool)
                 return output.strip()
```
ravisantoshgudimetla: While I like ease of use for sh module, it doesn't seem to be officially supported. So, I am wondering if it makes sense to use it or not. Let's have a vote on this and move forward as this is something that will be used again and again.

ims/einstein/ceph.py
```diff
@@ -230,34 +231,46 @@ def get_parent_info(self, img_id):
...
-        else:
+        # command = "echo {0} | sudo -S rbd --keyring {1} --id {2} unmap {3}".format(
+        #     self.password, self.keyring, self.rid, rbd_name)
+        # p = subprocess.Popen(command, shell=True,
+        #                      stderr=subprocess.STDOUT, stdout=subprocess.PIPE)
+        # output, err = p.communicate()
+        try:
+            with sh.sudo:
+                output = sh.rbd.unmap(rbd_name, keyring=self.keyring,
+                                      id=self.rid)
```
ravisantoshgudimetla: You may have to return output here.

ims/einstein/ceph.py
```diff
@@ -271,4 +284,5 @@ def showmapped(self):
                 maps[parts[2]] = parts[4]
             return maps
         else:
+            # Exception should be raised
```
ravisantoshgudimetla: Any specific reason for having a pass?

ims/einstein/hil.py
```diff
@@ -54,12 +54,12 @@ def resp_parse(self, obj):
...
-                raise haas_exceptions.AuthenticationFailedException()
+                raise netiso_exceptions.AuthenticationFailedException()
             elif obj.status_code == 403:
-                raise haas_exceptions.AuthorizationFailedException()
+                raise netiso_exceptions.AuthorizationFailedException()
             elif obj.status_code >= 400:
-                raise haas_exceptions.UnknownException(obj.status_code,
-                                                       obj.json()[
+                raise netiso_exceptions.UnknownException(obj.status_code,
+                                                         obj.json()[
```
ravisantoshgudimetla: space here..

ims/einstein/iscsi/iet/driver.py
```diff
@@ -1,33 +1,39 @@
...
         try:
             mappings = self.list_targets()
             if ceph_img_name in mappings:
-                raise iscsi_exceptions.NodeAlreadyInUseException()
+                raise iscsi_exceptions.TargetExistsException()
             rbd_name = self.fs.map(ceph_img_name)
             self.__add_mapping(ceph_img_name, rbd_name)
             self.restart_server()
-            self.__check_status(True)
+            # self.__check_status(True)
```
ravisantoshgudimetla: Remove unnecessary comments

ims/einstein/iscsi/iet/driver.py
```diff
@@ -123,8 +129,9 @@ def __remove_mapping(self, ceph_img_name, rbd_name):
 
     @log
     def restart_server(self):
-        command = "echo {0} | sudo -S service iscsitarget restart".format(
-            self.password)
+        # command = "echo {0} | sudo -S service iscsitarget restart".format(
+        #    self.password)
+        command = "service iscsitarget restart"
         p = subprocess.Popen(command, shell=True, stderr=subprocess.STDOUT,
```
ravisantoshgudimetla: I thought, we had to restart it as root, are you planning to add permissions to BMI user or is there any other change that won't need us to restart it as root.

ims/common/config.py
chemistry-sourabh: Felt that Network Isolator would be too big, so named it netiso. If you feel it gives more clarity I dont have an issue.

ims/common/config.py
chemistry-sourabh: 1. I am supposed to do that in config.py for the mandatory sections. The Other driver sections the respective drivers will verify.
2. You mean like get all the sections from the config file itself ? What if there is a spelling mistake in the section name ? You will still need the section name to add them to corresponding variables right ?

ims/database/db_connection.py
chemistry-sourabh: We didn't agree on it. I found while testing that the DB class is being loaded each time a request is coming, so as a side effect a new engine is being created each time. To do it the right way, there should have one engine and multiple sessions or else I fear we may not have consistency.

If you remember right you didn't want any sqlalchemy code in the operations.py or higher layer. So the only solution I see is creating an DBEngine Class that will be part of the RPC Server and a New DB Connection is passed to the BMI object when it is created. If we change sqlalchemy then we just need to change the DBEngine class without need for changing any other code.

ims/database/db_connection.py
chemistry-sourabh: This validation will happen in config.py

ims/einstein/iscsi/iet/driver.py
chemistry-sourabh: This is not unnecessary will need it again, will add a comment

ims/einstein/iscsi/iet/driver.py
chemistry-sourabh: I told to ignore IET for now

chemistry-sourabh: I am modifying it in a way such that there will a user called bmi who has sudo permissions and all files and folders will be under that user.

ims/einstein/ceph.py
chemistry-sourabh: ceph's unmap doesn't return anything

ims/einstein/ceph.py
chemistry-sourabh: As far as I know sh is not officially supported, but it is a very popular module. I was reluctant to use sh as it didn't have password for sudo support, but it was added recently and it works like a charm. Also the old code had shell=True without which it doesn't work and sh doesn't use it.

Also I checked the code sh doesn't use subprocess, but directly calls execv sys call.

chemistry-sourabh: @apoorvemohan @sirushtim @naved001 vote here

chemistry-sourabh: Also I can write another minimalistic wrapper that uses subprocess without shell=True to remove redundancy and give the same feel like sh.

ims/cli/cli.py
chemistry-sourabh: done

ims/cli/cli.py
chemistry-sourabh: I meant environment variable. I'll add an ENV in the name to make it more clear.

ims/common/config.py
chemistry-sourabh: Will keep that in mind when writing test cases

ims/common/constants.py
chemistry-sourabh: Done

ims/einstein/ceph.py
chemistry-sourabh: Updated comment

ims/einstein/ceph.py
chemistry-sourabh: Raising an exception and changed to try catch like others

ims/einstein/hil.py
chemistry-sourabh: This space is fine, it is pycharm formatting as per PEP8

ims/common/config.py
chemistry-sourabh: Do you want me to the validation in this PR ? Initially thought of doing it in Future PR.

chemistry-sourabh:
Pushed All changes, except the ones in dispute.

chemistry-sourabh:
@ravisantoshgudimetla check out the way I am creating ISCSI obj in operations.py. This may need to be discussed. I took some inspiration from HIL's implementation.

ravisantoshgudimetla:
@chemistry-sourabh , I am half done with reviewibg this PR, as it too big. First thing I would suggest is to split this PR, as it becomes easier for 2 reviewer as well. The idea is to make sure that Gene, Naved and others can also start reviewing.

ims/einstein/ceph.py
ravisantoshgudimetla: Then, you may have to return something to user saying that its mapped or else don't even bother saving it in output.

ravisantoshgudimetla:
Let me know, if you have some reservations against splitting the PR. I am not only talking in terms of numbers of files but also, number of changes - something like 
- mock drivers can be a separate PR altogether.
- Changes for reading config file can be a different PR.
- Sh changes.

chemistry-sourabh:
My only concern is how to split the files. Last time I split as per packages and you said you were lost. Technically even after the PR, BMI should still be runnable so it is difficult to split in such a manner such that BMI works after merging a PR.

ravisantoshgudimetla:
Usually a PR should be created based on changes. There should be one-to-one mapping between change and PR. So, it shouldn't be based on packages or anything.

naved001:
I think we should first make an issue where we can discuss possible solutions and then open a pull request to only solve that issue. This one is a huge pull request attempting to solve many issues at once.

chemistry-sourabh:
This PR has been successfully split so closing

PR #36, opened 2017-01-11
Delete bmi_client.py

chemistry-sourabh:
Deleting bmi_client.py as it was pushed by mistake.
We will be writing a client lib later, but I feel that this incomplete file will cause confusions.
So can keep it if people find a strong reason to keep it.


naved001:
I don't have a strong opinion on this. so +1

sirushtim:
+1

ravisantoshgudimetla:
I was under the impression, that this was used by other teams like Bigdata etc. Isn't it?

chemistry-sourabh:
I wrote it for Rajul. Bigdata guys must be using the CLI.

Issue #27, opened 2017-01-10
Fix * imports


chemistry-sourabh:
Currently in the code base there are a lot of * imports.  
The Task is to remove all the * imports and replace them with the correct imports.
This will put the codebase closer to PEP8 compliance.

chemistry-sourabh:
Solved in #37, except for tests which will be solved in tests PR

chemistry-sourabh:
@naved001 add a resolved label

PR #37, opened 2017-01-11
Imports

chemistry-sourabh:
Issue #27


sirushtim:
Nice work, but it seems like there is still some more imports to fix in the codebase.

ims/einstein/iscsi/iet.py:from ims.interfaces.iscsi import *
scripts/einstein_server:from ims.rpc.server.name_server import *
scripts/einstein_server:from ims.rpc.server.rpc_server import *
tests/database/test_project.py:from ims.database.database import *
tests/database/test_image.py:from ims.database import *
tests/einstein/test_operations.py:from ims.database import *
tests/picasso/test_rest.py:from ims.database import *

ravisantoshgudimetla:
I think @chemistry-sourabh wants to ignore both tests and iet.py but scripts/einstein is something that has to be fixed.

ravisantoshgudimetla:
@chemistry-sourabh: Squash the commits as well, it will be difficult to debug later, if we have too many commits(in case we want to do a git bisect and find which commit caused the issue).

chemistry-sourabh:
Fixed all except tests (Ignore them for now as they will be fixed when doing tests)

sirushtim:
I agree with the idea of squashing commits for PR's. One commit, one logical change per PR. It makes it a lot easier to review too.

Not sure why we don't want to do fix imports in tests with this change, but otherwise, looks good to me.

chemistry-sourabh:
After a lot of rebases finally squashed everything and also got travis to work again. (Atleast mastered rebasing now)

chemistry-sourabh:
@sirushtim tests need to rewritten as they won't work now. Can fix everything when doing that

chemistry-sourabh:
@naved001 or @ravisantoshgudimetla can be the second reviewer

ravisantoshgudimetla:
+1 and merging.

sirushtim:
@chemistry-sourabh Ah I see. Cool. Thanks.

Issue #28, opened 2017-01-10
Change BMIConfig obj


chemistry-sourabh:
Currently the BMIConfig obj reads each field from the config and stores them as a variable.
The Task here is to change these variables to dictionaries that store the entire section.
Also Checking for mandatory fields needs to be added.
This will allow for BMI to use Drivers.

sirushtim:
I'd much rather use oslo.config[1] which enables us to deal with configuration files in a declarative fashion thus saving a lot of time. It automatically has support for checking of mandatory fields too AFAIR.

[1] https://github.com/openstack/oslo.config

chemistry-sourabh:
This would require everybody to learn oslo which I feel is a bad idea. Also oslo is an openstack project does it make sense to use it ? I dont see any real advantage right now.
The only logical change to config would be to convert into json, but this style is much easier to write.

sirushtim:
So, think about it. In the long term, keeping declarative config code would be easier to maintain than having imperative code that would need continuous testing.

Oslo.config although under the openstack umbrella is a python library that's project agnostic and designed to only deal with config files.

Adding a config option with Oslo.config and making it a required/mandatory option is as simple as adding a single line of code which is easy for beginners tbh. Compare that to your method of adding several lines of code and testing for a new config option and furthermore maintaining separate config handling code is only going to waste time in the future IMO.

ravisantoshgudimetla:
This is an interesting discussion and @sirushtim has a valid point but I think using oslo may be an overkill in this scenario. As far as I understood, it has broad set of usages and one of them is reading INI file, others include RBAC policy enforcement but I liked the declarative approach. Why can't we combine both and use yaml for reading configurations(which is much more declarative) and use pyyaml or someother library for reading. Again, this needs an internal discussion and I advise taking votes and proceed further.

chemistry-sourabh:
I kind of like the later approach, but I am kind of confused on the whole declarative thing. Can anyone explain or put some nice link ?

sirushtim:
@ravisantoshgudimetla So, oslo comprises of a set of libraries, each one handling a specific use case. For example, the oslo.config library deals with INI config files. The oslo.policy library deals with RBAC policy enforcement. The suggestion is to not use all the oslo libraries. It is only to use oslo.config here. FWIW, YAML sounds like a nice idea too but that would mean testing and maintaining separate configuration handling code.

[...]

PR #41, opened 2017-01-11
Config

chemistry-sourabh:
Issue #28


chemistry-sourabh:
@naved001 @sirushtim @ravisantoshgudimetla start reviewing this

chemistry-sourabh:
@naved001 @sirushtim @ravisantoshgudimetla why dont you guys review this so that we can merge it and we can raise a different PR if we decide to go declarative. If we change this PR then all future PRs may need to be changed again. (Which is painful!!)

bmi_config.cfg
```diff
@@ -37,16 +36,22 @@ rpc_server_port = <port of rpc server>
 
 # this section is for specifying tftp settings
 [tftp]
-pxelinux_url = <path to pxelinux.cfg>
-ipxe_url = <path to location of ipxe files>
+pxelinux_path = <path to pxelinux.cfg>
+ipxe_path = <path to location of ipxe files>
 
 # this section is for http config
+# Should think of a better name for this section
```
naved001: we could call this `http_config`.

chemistry-sourabh: Na too boring name. It is obvious that it is related to http config as the section name is http and it is a config file.
Need something that gives more clarity than http. If you will http is sufficient I am ok with it.

sirushtim: How about [api] because that's what the http ip and port are for?

chemistry-sourabh: I think rest_api will make more sense ?

sirushtim: Sounds good.

chemistry-sourabh:
As per your reply @sirushtim you said that add an option in one line. I am adding the check in one line, but you didnt like that so can you elaborate on that ?

sirushtim:
@chemistry-sourabh So, adding a config option in one line is orthogonal to adding a check in one line. Each config option to be added in a line of code should correlate to one config option in the bmiconfig.cfg file. That's the declarative way of going about parsing config files. I think I posted a link to the paradigm of declarative programming already. It'd be good to read that first.

But, here's what I'd like the library API for config parsing to be anyway.

Take for example we have a config option called tftp_url:

The line of code we will add for that is:
- ConfigOption(name='tftp_url', section='tftp')

If we needed it to be a required option, it'd be:

- ConfigOption(name='tftp_url', section='tftp', required=True)

If we needed it to have a default variable, it would be

- ConfigOption(name='tftp_url', section='tftp', default='127.0.0.1')

In the end, we'd have a single file with only ConfigOption's that will relate very closely to the configuration file itself. Something like:

ConfigOption(name='tftp_url', section='tftp')
ConfigOption(name='debug', section='logs', required=False)
...
...
ConfigOption(name='haas_url', section='haas', required=True)


Keeping our config parser this way is going to make it easy to read and easy to maintain and also super convenient for new developers to add new options in the long run. Furthermore, it's an easy way to generate sample config files.

What already exists is hard to maintain nested dictionaries as a config data structure, that you're then writing separate checks for which can easily lead to dead code.

YAML/INI file selection makes little difference to me at this point. As long we have easy maintainable code for new developers to easily pick up on, that seals the deal.

Let me know if you have questions/still need more explanation on this.

chemistry-sourabh:
I think I am getting it a bit now. I wanted to store certain sections in the form of objects (like iscsi ) so that it can be passed to the driver instead of passing the entire config. Any ideas on that ? the way I understand it is all these config options will add an attribute in the config object , but what about sections ?

ravisantoshgudimetla:
@chemistry-sourabh , I think @sirushtim has provided from excellent feedback on the way it should be done. I like your intention but not the approach. One simple way of doing this is to write a decorator after parsing happens. This should tell if its an mandatory option or not. Let me know, if you want me to provide some psuedo code.

ravisantoshgudimetla:
> the way I understand it is all these config options will add an attribute in the config object , but what about sections ? 

We can have a section parser as well similar to options.(This will be dictionary of commands and options dictionary).

chemistry-sourabh:
dictionary of commands ? what does that mean ? I understood how @sirushtim wants it, but I dont get how can a decorator be used to mark mandatory option

chemistry-sourabh:
@sirushtim @ravisantoshgudimetla knock yourselves out

sirushtim:
Excellent work @chemistry-sourabh . The code looks so much cleaner now. Just some more changes and we're all good to go with this from my side.

Separate the parsing bits into it's own file. Keep all the config options declared in a separate python file, this will be an easy go to file to add/replace/remove config options file and the parser can be a separate library that can be imported/unit-tested.

It would also be a little easier for reviewers if the commit was separated into the following two PR's:

- Change in config to declarative style.
- Change in config keys.
 
Since the two solve different problems.

Also, we shouldn't be checking for cfg.value == 'True' in the code. Instead, we should sanitize the value to a boolean while parsing the config file.

chemistry-sourabh:
I am fine with everything except splitting the PR. If it is split then BMI wont work due to inconsistencies.

ims/picasso/rest.py
```diff
@@ -23,7 +23,8 @@ def setup_rpc():
 def start():
     setup_rpc()
     cfg = config.get()
-    app.run(host=cfg.bind_ip, port=cfg.bind_port)
+    app.run(host=cfg.rest_api.ip,
+            port=int(cfg.rest_api.port))
```
sirushtim: Would be good to validate this while parsing the config file. That way, we can keep delegate all type validation to the config parser.

chemistry-sourabh:
Used some of my PDP skills to make it pretty. @sirushtim @ravisantoshgudimetla

ims/common/bmi_config.py
```diff
@@ -0,0 +1,43 @@
...
+    cfg.option(constants.TFTP_SECTION, constants.IPXE_PATH_OPT)
+
+    # REST API Section
+    cfg.option(constants.REST_API_SECTION, constants.REST_API_IP_OPT)
+    cfg.option(constants.REST_API_SECTION, constants.REST_API_PORT_OPT, int)
+
+    # Logs Section
+    cfg.option(constants.LOGS_SECTION, constants.LOGS_PATH_OPT)
+    cfg.option(constants.LOGS_SECTION, constants.LOGS_DEBUG_OPT, bool)
+    cfg.option(constants.LOGS_SECTION, constants.LOGS_VERBOSE_OPT, bool)
```
sirushtim: We should make the data type a keyword arg or otherwise it'll be hard to change the API of the config parser in the future.

So this will be, 
cfg.option(constants.LOGS_SECTION, constants.LOGS_VERBOSE_OPT, type=bool)

chemistry-sourabh: keyword arg is optional. Do you want me to use it everywhere ? I am afraid it will cross the line length.

sirushtim: Yeah, if we don't use keyword arg, then we risk being stuck to a certain library API with little flexibility in the future.

That's true, we will exceed the line length shown below and it may look a little less prettier but I don't think we should lose API flexibility for that. We're eventually going to be adding more options on a case by case basis. Like one idea was to validate if the IP address/Port in the config file was reachable. Another was we could validate if the Ceph Pool itself existed. If any of these fail, we throw a log.ERROR that some of the HW resources are not reachable and possibly send a notification to the administrator of BMI. This is up for discussion, would like more reviews on it but I'm digressing :P . Basically, API flexibility over prettiness for me though.  

cfg.option(constants.LOGS_SECTION, constants.LOGS_VERBOSE_OPT,
                 type=bool)

ravisantoshgudimetla:
@chemistry-sourabh , I will able to review this over the weekend. If you think that it needs immediate attention, you can get this PR reviewed by @naved001 or @apoorvemohan. Sorry about that.

chemistry-sourabh:
Added it @sirushtim. Read the rest of the PR also

chemistry-sourabh:
@sirushtim Are you still looking at this ?

ims/rpc/server/rpc_server.py
```diff
@@ -39,15 +39,16 @@ def remake_mappings(self):
 @log
 def start_rpc_server():
     cfg = config.get()
-    if cfg.is_service:
+    if cfg.bmi.service == 'True':
```
sirushtim: This should be testing against the boolean value.

chemistry-sourabh:
@sirushtim Done!

sirushtim:
All the concerns I had were taken care of. So +1 from me.

chemistry-sourabh:
Yeah!! I will probably cry now

chemistry-sourabh:
@naved001 or @ravisantoshgudimetla  whoever is more jobless

Issue #29, opened 2017-01-10
TGT


chemistry-sourabh:
Currently we have a working TGT Driver, but we need to
1. Change subprocess to sh or atleast a wrapper as there is a lot of duplicate code.
2. If are sticking with subprocess remove all the shell=True.
3. Use sudo without password which will require the user or installer to create a new user with NOPASSWD sudo access. This user will have the permissions to all the files required by bmi.

chemistry-sourabh:
I am thinking of writing a wrapper for subprocess instead of using sh as sh doesn't use subprocess anymore. And as Ravi mentioned in previous PR subprocess is officially supported

chemistry-sourabh:
PR is merged

naved001:
After discussion with @apoorvemohan  we decided to keep it merged.

PR #42, opened 2017-01-12
Tgt changes

chemistry-sourabh:
Issue #29

### DIFF
--- a/ims/common/shell.py
+++ b/ims/common/shell.py
@@ -1,7 +1,7 @@
 # This is a wrapper for subprocess that must be used by all shell calls
 # The sudo in this doesnt take password as we wont be needing it.
 # shell=True must never be used
-
+# TODO Add unittests
 import subprocess
 
 from ims.common.log import create_logger, trace
--- a/ims/common/shell.py
+++ b/ims/common/shell.py
@@ -5,8 +5,14 @@
 import subprocess
 
 
-# Calls Command
 def call(command, sudo=False):
+    """
+    Executes the given command in shell
+
+    :param command: the command to execute as string
+    :param sudo: whether to execute as root (default is False)
+    :return: output of command as string
+    """
     s_command = command.split()
     if sudo:
         s_command.insert(0, 'sudo')
--- a/ims/common/shell.py
+++ b/ims/common/shell.py
@@ -4,7 +4,13 @@
 
 import subprocess
 
+from ims.common.log import create_logger, trace
+from ims.exception import shell_exceptions
 
+logger = create_logger(__name__)
+
+
+@trace
 def call(command, sudo=False):
     """
     Executes the given command in shell
@@ -16,5 +22,46 @@ def call(command, sudo=False):
     s_command = command.split()
     if sudo:
         s_command.insert(0, 'sudo')
-    output = subprocess.check_output(s_command, stderr=subprocess.STDOUT)
+    try:
+        output = subprocess.check_output(s_command, stderr=subprocess.STDOUT)
+        return output
+    except subprocess.CalledProcessError as e:
+        raise shell_exceptions.CommandFailedException(str(e))
+
+
+@trace
+def call_service_command(command, service_name, final_status=None):
+    """
+    Calls the given service command and checks whether it has status afterwards
+
+    :param command: service command to call
+    :param service_name: The service name
+    :param final_status: The final status of the daemon after call
+    :return: Output of Call
+    """
+    full_command = "service %s %s" % (service_name, command)
+    output = call(full_command, sudo=True)
+    if final_status is not None:
+        status = get_service_status(service_name)
+        if status is not final_status:
+            raise shell_exceptions.ServiceCommandFailedException(status)
     return output
+
+
+@trace
+def get_service_status(service_name):
+    """
+    Returns the status of given service
+
+    :param service_name: the service whose status should be returned
+    :return: Running or Dead or Status String if Error
+    """
+    status_string = call_service_command('status', service_name)
+
+    if 'active (running)' in status_string:
+        return 'Running'
+    elif 'inactive (dead)' in status_string:
+        return 'Dead'
+    # Have to check if there are any other states
+    else:
+        return status_string
--- a/ims/common/shell.py
+++ b/ims/common/shell.py
@@ -1,0 +1,14 @@
+# This is a wrapper for subprocess that must be used by all shell calls
+# The sudo in this doesnt take password as we wont be needing it.
+# shell=True must never be used
+
+import subprocess
+
+
+# Calls Command
+def call(command, sudo=False):
+    s_command = command.split()
+    if sudo:
+        s_command.insert(0, 'sudo')
+    output = subprocess.check_output(s_command, stderr=subprocess.STDOUT)
+    return output
--- a/ims/einstein/iscsi/iet.py
+++ b/ims/einstein/iscsi/iet.py
@@ -13,7 +13,7 @@ from ims.interfaces.iscsi import ISCSI
 logger = create_logger(__name__)
 
 
-# Dont Use IET until fixed in future PR
+# TODO Should Fix IET Driver (Issue #31)
 class IET(ISCSI):
     @log
     def __init__(self, fs, password):
--- a/ims/einstein/iscsi/iet.py
+++ b/ims/einstein/iscsi/iet.py
@@ -13,6 +13,7 @@ from ims.interfaces.iscsi import ISCSI
 logger = create_logger(__name__)
 
 
+# Dont Use IET until fixed in future PR
 class IET(ISCSI):
     @log
     def __init__(self, fs, password):
--- a/ims/einstein/iscsi/tgt.py
+++ b/ims/einstein/iscsi/tgt.py
@@ -12,9 +12,7 @@ logger = create_logger(__name__)
 
 
 class TGT(ISCSI):
-    '''
-        Class for implementing TGT
-    '''
+    """ Class for implementing TGT """
 
     def __init__(self, fs_config_loc, fs_user, fs_pool):
         self.TGT_ISCSI_CONFIG = "/etc/tgt/conf.d/"
@@ -24,10 +22,11 @@ class TGT(ISCSI):
 
     @log
     def start_server(self):
-        '''
-        Have to parse the output and send a status code.
-        :return:
-        '''
+        """
+        Starts the tgt service
+
+        :return: None
+        """
         try:
             command = "service tgtd start"
             output = shell.call(command, sudo=True)
@@ -39,10 +38,11 @@ class TGT(ISCSI):
 
     @log
     def stop_server(self):
-        '''
-        Need to check if this
-        :return:
-        '''
+        """
+        Stops the tgtd service
+
+        :return: None
+        """
         try:
             command = "service tgtd stop"
             output = shell.call(command, sudo=True)
@@ -54,15 +54,21 @@ class TGT(ISCSI):
 
     @log
     def restart_server(self):
-        '''
-        Again have to parse the output and send the values.
-        :return:
-        '''
+        """
+        Restarts the tgtd service
+
+        :return: None
+        """
         self.stop_server()
         self.start_server()
 
     @log
     def show_status(self):
+        """
+        Returns the status of tgtd
+
+        :return: Running or Dead or Error as String
+        """
         command = "service tgtd status"
         status_string = shell.call(command, sudo=True)
         logger.debug("Output = %s", status_string)
@@ -72,9 +78,15 @@ class TGT(ISCSI):
             return 'Dead'
         # Have to check if there are any other states
         else:
-            return "Running in error state"
+            return "Error"
 
     def __generate_config_file(self, target_name):
+        """
+        Generates the config file in conf.d/
+
+        :param target_name: Target for which config file should be created
+        :return: None
+        """
         config = open(
             os.path.join(self.TGT_ISCSI_CONFIG, target_name + ".conf"), 'w')
         template_loc = os.path.abspath(
@@ -90,11 +102,12 @@ class TGT(ISCSI):
 
     @log
     def add_target(self, target_name):
-        '''
+        """
         Adds target
-        :param target_name: Name of target to be added.
-        :return:
-        '''
+
+        :param target_name: Name of target to be added
+        :return: None
+        """
         try:
             targets = self.list_targets()
             if target_name not in targets:
@@ -111,10 +124,12 @@ class TGT(ISCSI):
 
     @log
     def remove_target(self, target_name):
-        '''
-        Removes target specified. Same as comment for above function.
-        :return:
-        '''
+        """
+        Removes target specified
+
+        :param target_name: Name of target to be removed
+        :return: None
+        """
         try:
             targets = self.list_targets()
             if target_name in targets:
@@ -132,11 +147,11 @@ class TGT(ISCSI):
 
     @log
     def list_targets(self):
-        '''
-        Lists all the targets available. This queries tgt-admin and gets the
-        list of targets
-        :return:
-        '''
+        """
+        Lists all the targets available by querying tgt-admin
+
+        :return: None
+        """
         try:
             command = "tgt-admin -s"
             output = shell.call(command, sudo=True)
--- a/ims/einstein/iscsi/tgt.py
+++ b/ims/einstein/iscsi/tgt.py
@@ -3,6 +3,7 @@ import subprocess
 import os
 import re
 
+import ims.common.shell as shell
 import ims.exception.iscsi_exceptions as iscsi_exceptions
 from ims.common.log import create_logger, log
 from ims.interfaces.iscsi import ISCSI
@@ -15,16 +16,11 @@ class TGT(ISCSI):
         Class for implementing TGT
     '''
 
-    # We should get all the config related to ceph in here.
-    # Also, root_password is something that can be avoided.
-
-    # Also removed sudo everywhere since it was causing issues
-    def __init__(self, fs_config_loc, fs_user, root_password):
-        self.arglist = ["service", "tgtd"]
+    def __init__(self, fs_config_loc, fs_user, fs_pool):
         self.TGT_ISCSI_CONFIG = "/etc/tgt/conf.d/"
         self.fs_config_loc = fs_config_loc
         self.fs_user = fs_user
-        self.root_password = root_password
+        self.fs_pool = fs_pool
 
     @log
     def start_server(self):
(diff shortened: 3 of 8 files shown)


chemistry-sourabh:
Rebased this have look @CCI-MOC/bmi-breakers

apoorvemohan:
@ravisantoshgudimetla can you please review this PR over the weekend (we need this for the march 15 deadline).

ravisantoshgudimetla:
@chemistry-sourabh , As discussed in other meeting earlier, please have valid docstrings PEP8 and 257 validations in place or else we will end up having dev branch similar to master.

ravisantoshgudimetla:
@chemistry-sourabh I am done with my first pass.

chemistry-sourabh:
so you approve ?

chemistry-sourabh:
@ravisantoshgudimetla do you want me to push TGT Driver changes to this PR as well ?

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
+
+        :return: None
+        """
+        try:
+            command = "service tgtd stop"
+            output = shell.call(command, sudo=True)
+            logger.debug("Output = %s", output)
+            if self.show_status() is not 'Dead':
+                raise iscsi_exceptions.StopFailedException()
+        except subprocess.CalledProcessError as e:
```
chemistry-sourabh: I dont think so PEP8 would have failed otherwise

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
+        status_string = shell.call(command, sudo=True)
+        logger.debug("Output = %s", status_string)
         if 'active (running)' in status_string:
             return 'Running'
         elif 'inactive (dead)' in status_string:
             return 'Dead'
         # Have to check if there are any other states
         else:
-            return "Running in error state"
+            return "Error"
```
chemistry-sourabh: We are not using this String anywhere in the errors. Also It is a state not what the error is. States should always be single words.

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
+        :return: None
+        """
+        config = open(
+            os.path.join(self.TGT_ISCSI_CONFIG, target_name + ".conf"), 'w')
         template_loc = os.path.abspath(
             os.path.join(os.path.dirname(os.path.abspath(__file__)), "..",
                          ".."))
-        for line in open(template_loc + "/tgt_target.temp", 'r'):
+        for line in open(os.path.join(template_loc, "tgt_target.temp"), 'r'):
             line = line.replace('${target_name}', target_name)
```
chemistry-sourabh: You mean ', ' ? That is pep8

ims/einstein/iscsi/iet.py
```diff
@@ -13,6 +13,7 @@
 logger = create_logger(__name__)
 
 
+# Dont Use IET until fixed in future PR
```
chemistry-sourabh: Issue is raised for this. You want me to add a TODO in the comment ?

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
-        command = "service tgtd start"
-        proc = subprocess.Popen(command, shell=True, stdout=subprocess.PIPE,
-                                stderr=subprocess.PIPE)
-        output, err = proc.communicate(self.root_password +
-                                       "\n")
-        logger.debug("Output = %s, Error = %s", output, err)
-        if self.show_status() is 'Running':
+        """
+        Starts the tgt service
+
```
chemistry-sourabh: Pycharm generated it this way, so I guess this is how majority of the people do it. Also I guess this is the format supported by the pydoc tool which generates docs from docstrings (Got to check)

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
+        :return: None
+        """
+        try:
+            command = "service tgtd start"
```
chemistry-sourabh: The name shouldn't change and I feel it is a bad idea to keep everything in the config, unless you have a better way to handle this

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
+    """ Class for implementing TGT """
 
-    # We should get all the config related to ceph in here.
-    # Also, root_password is something that can be avoided.
-
-    # Also removed sudo everywhere since it was causing issues
-    def __init__(self, fs_config_loc, fs_user, root_password):
-        self.arglist = ["service", "tgtd"]
+    def __init__(self, fs_config_loc, fs_user, fs_pool):
         self.TGT_ISCSI_CONFIG = "/etc/tgt/conf.d/"
```
chemistry-sourabh: This can be handled by pushing the TGT Driver code into this PR. Do you want me to ?

ims/tgt_target.temp
```diff
@@ -1,7 +1,7 @@
 <target ${target_name}>
     driver iscsi
     bs-type rbd
-    backing-store boot-disk-prototype/${target_name}
+    backing-store ${pool}/${target_name}
     bsopts ""conf=${ceph_config};id=${ceph_user}""
     write-cache off
     initiator-address ALL
```
chemistry-sourabh: I feel tgt_ceph_target will make more sense

ims/exception/iscsi_exceptions.py
```diff
@@ -77,10 +57,11 @@ def __init__(self, error):
...
     def __str__(self):
-        return "Failed to update config with error " + self.error
+        return "List Target Failed with error {0}".format(self.error)
 
 
 class StopFailedException(ISCSIException):
+    """ Should be raised when stopping iscsi failed """
     @property
     def status_code(self):
         return 500
```
chemistry-sourabh: I agree this needs to be debated on

ims/einstein/iscsi/iet.py
ravisantoshgudimetla: Yes..

ims/einstein/iscsi/tgt.py
ravisantoshgudimetla: No, if you want to handle this separately, put a TODO with issue number and when you are handling that, remove it..

ims/einstein/iscsi/tgt.py
ravisantoshgudimetla: Yeah, just stick to something which is clear as of now.

ims/einstein/iscsi/tgt.py
ravisantoshgudimetla: So exception is a state and its always good to know why it has gone to error state. I mean, the path in state diagram is important too as we need to find why this has gone to a particular error state.

ims/einstein/iscsi/tgt.py
ravisantoshgudimetla: No, I was asking about the green highlight for patch. Not sure, if its something to show difference though.

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
-            raise iscsi_exceptions.UpdateConfigFailedException(e.message)
-        except (iscsi_exceptions.MountException,
-                iscsi_exceptions.DuplicatesException) as e:
-            # Do we still need the above exceptions? Duplicate
-            # exception can be avoided by checking lists as shown above
-            # Mouting is something that we are not doing now?
-            logger.info("Error exposing iscsi_target")
-            raise e
+            raise iscsi_exceptions.TargetDeletionFailed(str(e))
+        except subprocess.CalledProcessError as e:
```
ravisantoshgudimetla: I guess, you need to address this.

ims/tgt_target.temp
ravisantoshgudimetla: I had something like tgt could be backed by anything like ceph or gluster or someother filesystem so, somewhere backend should be specified is what I thought.

ims/einstein/iscsi/iet.py
chemistry-sourabh: ok

ims/einstein/iscsi/tgt.py
chemistry-sourabh: ok

ims/tgt_target.temp
chemistry-sourabh: ok not a huge issue

ims/einstein/iscsi/tgt.py
chemistry-sourabh: Now I remember I initially put the try catch in shell.call(), then I needed to pass the exception class to the function which sort of felt ridiculous. So I caught it in each function, although it feels duplicated, it is much more readable I feel.

ims/einstein/iscsi/tgt.py
chemistry-sourabh: I agree with you, but show status cant tell that. The one calling show_status handles the raising of the required exception. For example in start_server() it checks status and if is not running then it raises StartFailedException

chemistry-sourabh:
@knikolla please review this

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
-                command = "tgt-admin --execute".format(self.root_password)
-                proc = subprocess.Popen(command, shell=True,
-                                        stdout=subprocess.PIPE,
-                                        stderr=subprocess.PIPE)
-                proc.communicate()
-                if proc.returncode != 0:
-                    raise iscsi_exceptions.UpdateConfigFailedException(
-                        "Adding new target failed at creating target file \
-                        stage")
+                command = "tgt-admin --execute"
```
knikolla: The ``--execute`` part of the command doesn't need to be in the config file, here it's fine. The executable to run the command on though (eg. tgt-admin), must be configurable.

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
+        :return: None
+        """
         try:
             targets = self.list_targets()
             if target_name in targets:
-                os.remove(self.TGT_ISCSI_CONFIG + target_name + ".conf")
-                # tgtarglist = ["sudo", "-S","tgt-admin", "--execute"]
+                os.remove(os.path.join(self.TGT_ISCSI_CONFIG,
+                                       target_name + ".conf"))
                 command = "tgt-admin -f --delete {0}".format(target_name)
```
knikolla: Same comment about ``tgt-admin`` as above.

ims/einstein/iscsi/tgt.py
knikolla: While I agree that the exception to be caught here shouldn't be ``subprocess.CalledProcessError``, you would still need to raise another exception in ``shell.call()`` and catch that one here.

ims/einstein/iscsi/tgt.py
```diff
@@ -11,180 +12,154 @@
...
-            # This exception has to be modified. I prefer writing a new
-            # exception which is something like iscsi_communication exception
-        else:
+        """
+        Lists all the targets available by querying tgt-admin
+
+        :return: None
+        """
+        try:
+            command = "tgt-admin -s"
```
knikolla: Same comment about ``tgt-admin``. Make it configurable.

ims/einstein/iscsi/tgt.py
chemistry-sourabh: ok this seems reasonable. But I have two questions @knikolla 

1. What if in the future people whip up some new service management utility like systemctl or service that will require a code change right ?
2. What if I want to handle my statuses in my own way in the driver like raise specific exceptions for error messages in the status ? do I add a get_status method also ?

@ravisantoshgudimetla what you think about this ?

ims/common/shell.py
```diff
@@ -0,0 +1,20 @@
+# This is a wrapper for subprocess that must be used by all shell calls
+# The sudo in this doesnt take password as we wont be needing it.
+# shell=True must never be used
```
chemistry-sourabh: The reason I put it so that other new developers do not use shell=True by mistake

ims/einstein/iscsi/tgt.py
chemistry-sourabh: It makes sense to raise an exception if start or stop of a service failed. This method is supposed to just return the status of the tgtd service, so I feel it doesn't make sense to raise an exception unless service tgtd status to execute. I agree it makes sense to add it in shell.py

chemistry-sourabh:
Also @ravisantoshgudimetla do we still start_server, ... in the interface. It clearly depends on the driver and we are not calling it anywhere higher up in operations

ims/einstein/iscsi/tgt.py
knikolla: 1. If people whip up some new service management utility, than you don't need to update the driver code anymore because you have abstracted away the pieces the driver doesn't care about. The driver cares about "a way" to restart the service, and ``shell.py`` provides it. If you have N drivers, you now don't need to change ``service blabla restart`` in N drivers, but only once in ``shell.restart_service``.
2. Yes. The driver will ask to restart the service, ``shell.py`` will fail restarting and will raise and exception, the driver will catch that exception and ``check_status()`` and either raise an exception which may be more driver specific, or handle that exception gracefully.

chemistry-sourabh: So I am guessing the exception contains the output of the "service status" in the message ?

knikolla: Yes.

ravisantoshgudimetla: @knikolla IMO, start_server & stop_server are functions specific to an iscsi server and I think, it makes sense to have them in that particular file rather than in shell.py. To me, shell.py is a util file that each driver could have. If there is an iscsi server that needs to be started from python or some other language, having it in its own set of scripts, having a generic one may not work either. 

I understand your point of abstracting shell code away but I think this is cleaner we don't know, what iscsi server uses what. Let me know what you think.

ims/einstein/iscsi/tgt.py
ravisantoshgudimetla: I agree with @knikolla.

chemistry-sourabh:
@knikolla and @ravisantoshgudimetla please settle on the start server, stop server changes so that I can make the required changes.

ims/einstein/iscsi/tgt.py
knikolla: @ravisantoshgudimetla @chemistry-sourabh Alright, since there's only really one server you need to start, it's ok like this. But, once you have more drivers/services to start, judge the amount of code you can reuse and refactor it in a common module.

knikolla:
@chemistry-sourabh Commented on the start server discussion.

ims/einstein/iscsi/tgt.py
chemistry-sourabh: @ravisantoshgudimetla the way I see it even IET needs to do a service restart, so it is better to go with @knikolla's approach and call the start_service, etc functions in the iscsi driver.

ravisantoshgudimetla: I am not against moving this part but what I am telling is starting and stopping a server are logically part of that iscsi server, so if we move them to shell.py, what if there is an iscsi server that starts and stops using some Python script, should we have another Python.py for that or wrap it in shell.py? I think I am OK with moving as long as start and stop server are available as methods in TGT or iscsi but call the ones in shell.py or implement what @knikolla suggested for now.

chemistry-sourabh: The way I understood it is the interface methods will stay in TGT, but the logic of starting and stopping servers will be moved to shell.py and we will call these functions in TGT's interface methods. Correct me if I am wrong @knikolla

chemistry-sourabh:
@knikolla and @ravisantoshgudimetla have a look!

ims/exception/iscsi_exceptions.py
```diff
@@ -82,9 +89,21 @@ def __str__(self):
...
+
     @property
     def status_code(self):
         return 500
 
     def __str__(self):
         return "ISCSI Failed to Start"
+
+
+class CheckStatusFailed(ISCSIException):
```
knikolla: Doesn't follow the naming standard.

chemistry-sourabh:
Yes the todos are more related to config changes which will be done in TGT Driver.

ims/exception/iscsi_exceptions.py
chemistry-sourabh: Will change it to ShowStatusFailed

chemistry-sourabh:
@knikolla have a look

chemistry-sourabh:
Thanks @knikolla

ims/common/shell.py
```diff
@@ -0,0 +1,67 @@
...
+    """
+    status_string = call_service_command('status', service_name)
+
+    if 'active (running)' in status_string:
+        return 'Running'
+    elif 'inactive (dead)' in status_string:
+        return 'Dead'
+    # Have to check if there are any other states
+    else:
+        return status_string
```
ravisantoshgudimetla: I am OK with this file. Can you add a to-do saying that this needs unit tests.

Rest LGTM.

chemistry-sourabh:
Done!! @ravisantoshgudimetla

ravisantoshgudimetla:
squash the commits and we can merge..

chemistry-sourabh:
Approve it and we can squash it from GitHub only

naved001:
Before we merge this, has anybody tested this with real hardware?

apoorvemohan:
@naved001 Can you deploy it on NEU-dev and test it?

chemistry-sourabh:
I tested on this my dev before submitting. Only difference is my HIL uses mock nodes.

naved001:
@apoorvemohan @chemistry-sourabh works with real nodes!

ravisantoshgudimetla:
@naved001 Merge it, if you are ok and squash the commits before merging.

apoorvemohan:
@naved: This should be merged after we have the it passes the new test cases that you and @Kan are working on. Please unmerge this.

naved001:
@apoorvemohan The tests that I am writing are unittests, and those do no test the TGT driver. We have tested this manually with real hardware so I thought that was sufficient. If you still insist, I can unmerge it. 

Moreover, in future, if you would like @CCI-MOC/bmi-breakers to not merge PRs for whatever reasons, please either put a 'DO NOT MERGE YET' label or a red cross.

chemistry-sourabh:
Red cross would be a bad idea. @apoorvemohan let it be, this was blocking other tests PRs. @naved001 also successfully tested this on real hardware. Besides for the known issues this code doesnt create any new issues

apoorvemohan:
Had a discussion with @naved001. No need to unmerge it.

Issue #53, opened 2017-01-18
Incorporate DB Init/Migration tool


sirushtim:
Incorporate sqlalchemy-migrate/alembic into BMI in order to version database changes for upgrade. Furthermore, create a db-init script that will bootstrap BMI's database using the corresponding migration library.

chemistry-sourabh:
@ravisantoshgudimetla want to take a shot at this ? You are the official DB Admin anyways :P

PR #60, opened 2017-02-06
BMI Install script

sirushtim:
The scripts add a basic BMI installation with the following assumptions:

- Mock HIL
- Atleast 5G free hard drive space
- Ubuntu or RHEL Distro
- sudo access


scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+cp examples/haas.cfg.dev-no-hardware ./haas.cfg
+
+export HAAS_ENDPOINT=http://127.0.0.1:5000
+export HAAS_USERNAME=haas
+export HAAS_PASSWORD=secret
+
+# Start Haas services
+haas-admin db create
+haas serve 5000&
+haas serve_networks&
```
naved001: AFAIK, we don't need to run `haas serve_networks` if you are doing `haas serve <port>`.

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+### START BMI setup
+git clone -b dev http://github.com/cci-moc/ims
+pushd ims
+
+### Install BMI
+
+sudo mkdir -p /var/log/bmi/logs /etc/bmi /opt/bmi
+sudo chown -R ubuntu:ubuntu /etc/bmi /var/log/bmi /opt/bmi
+
+wget https://gist.githubusercontent.com/sirushtim/6c28b7168f9121284364e46476af0c73/raw/a9f47dbf35a891689bfc9fe34f5fdf89f6207550/bmi_config.cfg.test
```
naved001: hardcoded path to your own repo? Maybe we should add the this test cfg file to the main repo (in an `/examples` folder in the main repo).

naved001:
If I understand correctly, this 
* installs HIL for you
* sets up a mock node and mock switches. 
* installs ceph
* installs BMI and runs the `einstein` and `picasso` servers. 
* uploads a cirros image to ceph. 

If this script is to just install BMI, why does it make mock switches and nodes? Or are we doing both here, install BMI and test it with a fake setup?

Feel free to correct me if I misunderstood something here.

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+set -ex
+sudo apt-get install -y python-pip
+### Packages for BMI
+sudo apt-get install -y tgt sqlite3 virtinst
+
+git clone http://github.com/cci-moc/hil
+
+pushd hil/
+sudo apt-get install -y libvirt-bin bridge-utils ipmitool telnet apache2 libapache2-mod-wsgi python-pip qemu-kvm python-libvirt python-psycopg2 vlan net-tools icu-devtools libicu-dev libxml2-dev libxslt1-dev python-virtualenv python3-virtualenv virtualenv # For Ubuntu
+
```
naved001: Maybe we could do `cat /etc/*-release` to get the name of distro and based on it install the required packages. Because we are using CentOS (or RHEL) almost everywhere.

naved001:
Also, making the script just for ubuntu strikes me as a bad idea because we are using CentOS (or RHEL) almost everywhere. 

I think it would easy to change `install_bmi.sh` to support centOS and RHEL. But I am not sure about setting up ceph though (Are we setting up ceph just for the testing environment?)

chemistry-sourabh:
I advise separating HIL installation into a separate file and calling it in install-bmi.sh. You can do the same for install-ceph.sh also. If you are worried about HIL or Ceph already being installed you can make it interactive.

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+
+### START BMI setup
+git clone -b dev http://github.com/cci-moc/ims
+pushd ims
+
+### Install BMI
+
+sudo mkdir -p /var/log/bmi/logs /etc/bmi /opt/bmi
+sudo chown -R ubuntu:ubuntu /etc/bmi /var/log/bmi /opt/bmi
```
chemistry-sourabh: Create a bmi user and chown everything to that user. Also add that user to sudo group also

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+
+wget https://gist.githubusercontent.com/sirushtim/6c28b7168f9121284364e46476af0c73/raw/a9f47dbf35a891689bfc9fe34f5fdf89f6207550/bmi_config.cfg.test
+cp bmi_config.cfg.test /etc/bmi/bmiconfig.cfg
+cp ims/ipxe.temp /etc/bmi/ipxe.config
+cp ims/mac.temp /etc/bmi/pxelinux.cfg
+
+sudo python setup.py develop
+touch /opt/bmi/bmi.db
+bmi db ls
```
chemistry-sourabh: why did you ls here ?

scripts/install_bmi.sh
chemistry-sourabh: Can you wrap around long lines to 80 characters please

scripts/install_bmi.sh
sirushtim: Just followed the docs here: https://github.com/CCI-MOC/hil/blob/master/docs/INSTALL-devel.rst#start-the-server

I could just remove the line if you think it's of no use.

scripts/install_bmi.sh
sirushtim: Yes, this will be done in the next iteration. Thanks.

scripts/install_bmi.sh
sirushtim: This bootstraps the database with the necessary tables. Something https://github.com/CCI-MOC/ims/issues/53 intends to fix.

sirushtim:
@naved001 

> If I understand correctly, this
> installs  HIL for you
> sets  up a mock node and mock switches.
> installs ceph
> installs BMI and runs the einstein and picasso servers.
> uploads a cirros image to ceph.
> If this script is to just install BMI, why does it make mock switches and nodes? Or are we doing both here, install BMI and test it with a fake setup?

> Feel free to correct me if I misunderstood something here.

That sounds about right. Perhaps, the name install_bmi.sh is misleading. We can split setting up of HIL into its own script which would make it cleaner.

@naved001 We don't need to keep this script generic for all distros. This script isn't intended to deploy BMI into production. It is only going to be used for Development/Continuous Integration. However, improvements can be made in a later patch once we get our CI up and running.

@chemistry-sourabh Sure, splitting HIL into it's own script sounds good.

At the moment, getting `bmi pro` to run is a priority. That would serve us as an integration test in our CI to then satisfy the criteria to merge dev branch to master branch.

scripts/install_bmi.sh
naved001: you are right, I just learnt it today too. smh

ravisantoshgudimetla:
@sirushtim I recommend using ansible or some other automation tool for installation scripts, you can refer to haas ansible installation scripts by Kristi. The advantage is you don't have to take care of error handling stuff.

sirushtim:
@ravisantoshgudimetla This script is only used for development/CI testing. I'd prefer we keep this to BASH for now as it becomes really simple for newcomers to fix a bug that breaks our CI. I like Ansible for it's idempotent nature but Ansible feels like an over-optimization at this point for a few bash commands to run our CI.

However, for production, we should definitely create a new Ansible or a Puppet module for BMI thereby handling other infrastructure specific errors that may occur.

ravisantoshgudimetla:
@sirushtim My only concern is it will be a throw away work. I have some script in ansible that does most installation but I lost it somewhere. I will try digging or writing something by end of next week and send.

sirushtim:
@ravisantoshgudimetla IMO, a script to set up a development environment is not going to be throw away work. Each developer is expected to run the script to setup a basic BMI dev env quickly since we set up Ceph/HIL services which would take otherwise more time to write through Ansible playbooks. So basically, BASH is a super quick way to fix failing CI builds than Ansible is all.

However, if you do have Ansible scripts to set up BMI, it would be really nice to have in the repository as there are plans to use Ansible/Puppet when BMI when goes into production.

scripts/install_bmi.sh
chemistry-sourabh: Perhaps a better way to do it is to add a separate method that simply adds the default admin into the admin. I think HIL also has a command which does it, correct me if I am wrong @naved001

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+
+### Install BMI
+
+sudo mkdir -p /var/log/bmi/logs /etc/bmi /opt/bmi
+sudo chown -R ubuntu:ubuntu /etc/bmi /var/log/bmi /opt/bmi
+
+wget https://gist.githubusercontent.com/sirushtim/6c28b7168f9121284364e46476af0c73/raw/a9f47dbf35a891689bfc9fe34f5fdf89f6207550/bmi_config.cfg.test
+cp bmi_config.cfg.test /etc/bmi/bmiconfig.cfg
+cp ims/ipxe.temp /etc/bmi/ipxe.config
+cp ims/mac.temp /etc/bmi/pxelinux.cfg
```
chemistry-sourabh: There is no point of copying the template files. setup.py is going to install them along with other modules

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+
+sudo python setup.py develop
+touch /opt/bmi/bmi.db
+bmi db ls
+sqlite3 /opt/bmi/bmi.db "insert into project values (0, 'bmi_infra', 'bmi_network')"
+
+# Temporary changes to get stable CI. Need to do pull request to get this fixed.
+sed -i s/"app.run(host=cfg.bind_ip, port=cfg.bind_port)"/"app.run(host=cfg.bind_ip, port=int(cfg.bind_port))"/ ims/picasso/rest.py
+sed -i s/"HAAS_BMI_CHANNEL = \"vlan\/native\""/"HAAS_BMI_CHANNEL = \"null\""/ ims/common/constants.py
+sed -i s/"cluster = rados.Rados(rados_id=self.rid, conffile=self.r_conf)"/"cluster = rados.Rados(conffile=self.r_conf)"/ ims/einstein/ceph.py
```
chemistry-sourabh: Can you please elaborate why you need these 3 lines ?

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+sudo python scripts/einstein_server &
+sudo python scripts/picasso_server &
+
+sleep 5
+
+sudo chmod 777 /etc/tgt/conf.d/
+
+popd
+
+bmi db ls
```
chemistry-sourabh: why do you need a second one ?

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+sqlite3 /opt/bmi/bmi.db "insert into project values (0, 'bmi_infra', 'bmi_network')"
+
+# Temporary changes to get stable CI. Need to do pull request to get this fixed.
+sed -i s/"app.run(host=cfg.bind_ip, port=cfg.bind_port)"/"app.run(host=cfg.bind_ip, port=int(cfg.bind_port))"/ ims/picasso/rest.py
+sed -i s/"HAAS_BMI_CHANNEL = \"vlan\/native\""/"HAAS_BMI_CHANNEL = \"null\""/ ims/common/constants.py
+sed -i s/"cluster = rados.Rados(rados_id=self.rid, conffile=self.r_conf)"/"cluster = rados.Rados(conffile=self.r_conf)"/ ims/einstein/ceph.py
+
+sudo python scripts/einstein_server &
+sudo python scripts/picasso_server &
```
chemistry-sourabh: Don't start picasso or einstein here

scripts/install_bmi.sh
chemistry-sourabh: Also there is no need to touch the db as it will be created by the ls.

chemistry-sourabh:
@sirushtim I advise separating the script into installation and setup. Like install-bmi.sh will only install bmi and setup-bmi.sh will only configure it (like adding user and image) similarly for HIL and Ceph also.
Then you can make a central script that calls all of them individually ( Can make it interactive later if not replaced by ansible ).

Also make separate start-env and stop-env scripts that will run HIL, Ceph and TGT and stop them.
I feel this will make stuff modular and easy to maintain. Also add some echo statements so that it will be easy to identify where everything screwed up.

Also are you installing TGT ? If no, you should do it in a separate script ( like in first para )

chemistry-sourabh:
@ravisantoshgudimetla please review the ceph.sh file

chemistry-sourabh:
Thats all folks!!

scripts/install_bmi.sh
sirushtim: @chemistry-sourabh That better method to initiate database tables is what #53 was raised for. Please have a read at that bug for more information. 

There is a need to touch the db file since we want to retain file ownership of the script to be the one that is run by the script than the one that is invoked by einstein (example: root).

scripts/install_bmi.sh
sirushtim: No, at the moment, setup.py does not copy config/template files to /etc.

scripts/install_bmi.sh
sirushtim: These 3 lines are basically bugs that I faced while setting up bmi. I'll have to remove it in the next iteration and make it part of the pull request.

scripts/install_bmi.sh
sirushtim: This is just to know if the basic command works or else the script exits if this command fails.

scripts/install_bmi.sh
sirushtim: Besides modularizing the code, is there any other difference in starting the scripts at this stage?

sirushtim:
@chemistry-sourabh Yes, since this is an early patch just to get reviews on the functionality. The modularizing is definitely going to happen. Please try to run the script for yourself and if possible, test BMI's behaviour after that. It would be good to get more comments on the functionality and uncover possible setup sequences I may have possibly missed out.

scripts/install_bmi.sh
chemistry-sourabh: This is supposed to be an installation script. The dev is supposed to run them when he made some changes and stuff like that. So remove those lines

scripts/install_bmi.sh
chemistry-sourabh: What issue is with the third line ?

scripts/install_bmi.sh
chemistry-sourabh: those files dont go to /etc/ they go into site-packages or dist-packages and they should be there also as they shouldnt be tweaked my anyone

chemistry-sourabh:
I will try running it later. But wont be able to run provision as TGT is missing.

In this PR add the bmi user as the entire code revolves around it. It is important to setup proper permissions and users so that we wont have any unexpected issues when moving to production.

scripts/install_bmi.sh
chemistry-sourabh: Also dont use the gist. Push the cfg.test into this PR also.

chemistry-sourabh:
Nearly forgot need to install service/systemctl scripts also. I made them a while back will email you if I find them.

scripts/install_bmi.sh
sirushtim: Specifying conf is enough rather that specifying the RID as the rid is included in the conf, the line was throwing an error that I'll raise an issue later for.

scripts/install_bmi.sh
sirushtim: IMO, It doesn't make much sense for them to go to dist-packages. These are template files that are changed on a case by case basis and in fact, will be tweaked by deployers of BMI.

Also, FWIW, we have code in BMI to specify locations to config files specifically.

scripts/install_bmi.sh
sirushtim: That makes sense, I think we should still run it but run it in a screen/tmux session. Perhaps, the word install_bmi.sh is misleading, that could be changed to bmi_dev_env.sh

sirushtim:
> I will try running it later. But wont be able to run provision as TGT is missing.

Please note that TGT is installed in the very beginning and the tgt commands work fine. The script seems fine with regards to satisfying dependencies, it provisions BMI but I just want to find more setup sequences that are missing. So it would be really helpful if you could provide comments once you're done running the script.

> In this PR add the bmi user as the entire code revolves around it. It is important to setup proper permissions and users so that we wont have any unexpected issues when moving to production.

It doesn't matter at all to add a bmi user as that's just a change in the naming convention and it really doesn't make a difference for a developer script, I hope it's clear that this is a developer script for developers. Feel free to point out what proper permissions and users are not setup and what issues you think we might face when moving to production.

> Nearly forgot need to install service/systemctl scripts also. I made them a while back will email you if I find them.

Not necessary again as developers will just manually invoke the script to debug them.

I want to reiterate that we'll never use this script for production, it's only a quick way for a developer to run BMI in his own Linux box. We'll use Ansible or Puppet in the future to actually deploy BMI.

So if we could get more comments after you guys have run the script and tested BMI yourselves, that would be really helpful.

scripts/install_bmi.sh
chemistry-sourabh: Not for the temp files. They are hardcoded if I remember right

scripts/install_bmi.sh
chemistry-sourabh: why ? He will most probably stop it anyways. You can add it in the separate start-env.sh script (which I spoke about in the below comment), but it would be wise to let the dev start bmi manually as he might be trying to start it in infinite different ways. I guess @naved001 agrees with me.

chemistry-sourabh:
Github interface sucks. It would be good to mark the unread comments. Getting lost

scripts/install_bmi.sh
sirushtim: So, take this template file for example:

#!ipxe
      set keep-san 1
      ifconf --configurator=dhcp net2
      sanboot --keep iscsi:${iscsi_ip}:tcp:3260:1:${target_name}
      boot

The ethernet interface and the port name in the above template can change depending on the deployment. What you're suggesting is stripping away this flexibility the deployers can take advantage of.

scripts/install_bmi.sh
sirushtim: Since this is a quick way to setup BMI. It's only natural to run BMI in the end. The developer is free to kill the process and run their own process later. It feels like you're assuming only the developers that know their way around BMI will only use this script. That's not true all the time. When new developers come in, we want them to have a working BMI setup with the processes running so that they can just use the CLI and get bootstrapped with the functionality of BMI. It's not intuitive enough to expect them to run the scripts in the code because they won't know where the processes are in the code until they read the code. 

So, again, modularization to separate the running of the process into a separate function will happen in the next iteration.

Mostly, this opinion is a manifestation of my recent experience with BMI in that I had to go through the codebase first before trying BMI itself to understand the way things work. It should always be the opposite.

chemistry-sourabh: I agree with you. But do the modularization in this PR as it it will be easy to read. I will run this script and let you

naved001:
I ran the script right now on an Ubuntu VM and here are my findings:

* After running the scripts from the `scripts` folder, it had cloned `hil` and `ims` repos there. Why do we clone the `ims` repo again? It's not really a big deal though. 
* The end user has to set the HAAS_USERNAME and HAAS_PASSWORD variables. We should either document it or do it in the script. 
* Ultimately I couldn't provision the VM; I got the following message: 
`Failed to update config with error Adding new target failed at creating target file                         stage
`

Overall I like this a lot, after running just 2 scripts we had the BMI CLI running. Good for simple tests and for new devs to get started.

sirushtim:
@naved001 Great. Thanks a lot for running it. Replies inline.

> After running the scripts from the scripts folder, it had cloned hil and ims repos there. Why do we clone the ims repo again? It's not really a big deal though.

??

> The end user has to set the HAAS_USERNAME and HAAS_PASSWORD variables. We should either document it or do it in the script.

Oh yes, I had overlooked this step, we should definitely try doing it in the script. Thanks.

> Ultimately I couldn't provision the VM; I got the following message:
Failed to update config with error Adding new target failed at creating target file stage

Yeah, this is the same error I get too. It feels like something to do with TGT and the way BMI deals with permissions. Still debugging it...

scripts/install_bmi.sh
```diff
@@ -0,0 +1,104 @@
...
+### Setup HaaS network
+haas network_create_simple bmi_network bmi_infra
+
+# FIN HaaS setup
+deactivate
+popd
+
+
+### START BMI setup
+git clone -b dev http://github.com/cci-moc/ims
```
naved001: sorry that I wasn't clear earlier. My question is why are we cloning the ims repo here? Because to get this script the user/dev would have already cloned the ims repo. Does that make sense?

sirushtim: Good catch. That makes sense. I'll remove it.

naved001:
@xukan you might wanna try this!

scripts/install_bmi.sh
chemistry-sourabh: Couldn't the user just download the script and run it ?
Also @sirushtim I guess are you installing BMI also in a virtualenv ? If yes is it the same virtualenv HIL is being installed in ?
I would prefer BMI be installed in a separate virtualenv if it makes sense.
Perhaps a smarter way to do it is to detect whether the dev cloned the repo or not and clone it accordingly.

chemistry-sourabh:
@sirushtim why dont you update the script with requested changes. I will try it today evening

sirushtim:
@chemistry-sourabh Please feel free to test the script and post your reviews. It would be a good idea to just update once we have everyone's comments. Even if the script is updated, the functionality of the script remains the same. So, it would be good if you could try it and post the possible missing setup sequences you find with the script.

scripts/install_bmi.sh
sirushtim: BMI is not installed in a virtualenv because of a system python package dependency(ie. rados). FWIW, I would prefer projects be in their own virtualenv's too.

The user would always have to clone the repository to run the script especially because we're modularizing the script soon. The user downloading all the scripts specifically/manually to run it would be an anti pattern we should discourage. There will be clear instructions to guide the user wanting to run this script.

chemistry-sourabh: Then add it in a virtualenv. I use an virtualenv for my own environment. This will allow multiple people to develop on the same VM or server if required.

chemistry-sourabh:
+ qemu-img create -f qcow2 bmi_virtual_node.qcow2 8G
Formatting 'bmi_virtual_node.qcow2', fmt=qcow2 size=8589934592 encryption=off cluster_size=65536 lazy_refcounts=off refcount_bits=16
+ sleep 10
+ sudo virt-install -n BMIVM --description 'Test BMI VM' --os-type=Linux --os-variant=rhel6 --ram=2048 --vcpus=2 --disk path=bmi_virtual_node.qcow2,bus=virtio,size=10 --pxe --network bridge:virbr0
WARNING  KVM acceleration not available, using 'qemu'
WARNING  Graphics requested but DISPLAY is not set. Not running virt-viewer.
WARNING  No console to launch for the guest, defaulting to --wait -1

Starting install...
ERROR    internal error: process exited while connecting to monitor: warning: TCG doesn't support requested feature: CPUID.01H:ECX.vmx [bit 5]
warning: TCG doesn't support requested feature: CPUID.01H:ECX.vmx [bit 5]
2017-02-16T03:13:31.682067Z qemu-system-x86_64: cannot set up guest memory 'pc.ram': Cannot allocate memory
Domain installation does not appear to have been successful.
If it was, you can restart your domain by running:
  virsh --connect qemu:///system start BMIVM
otherwise, please restart your installation.
++ awk '{ print $1 }'
++ grep running
++ sudo virsh list
+ vm_id=
++ awk '-F'\''' '{print $2}'
++ grep 'mac address'
++ sudo virsh dumpxml
error: command 'dumpxml' requires <domain> option
+ mac_address=
+ haas node_register_nic bmi_node bmi_port
Invalid arguements.  Usage:
  node_register_nic <node> <nic> <macaddr>
      
    Register existence of a <nic> with the given <macaddr> on the given <node>

I guess this is because I was not running ceph. Better run the ceph script in the bmi script and run it also

sirushtim:
Nono, this is because the VM you were running this script on didn't have enough memory to allocate to a new Virtual Machine.

Notice this line:

> 2017-02-16T03:13:31.682067Z qemu-system-x86_64: cannot set up guest memory 'pc.ram': Cannot allocate memory

Try increasing the memory of the VM and run the script again. This brings up a good point in that we can add to the documentation of how much memory we'd need for this script to run successfully. 

Also, when the modularization does happen, I'll have a script that runs everything else at one go. For now, it'd be good to run the ceph script before running the bmi script. The two hardly take any time. ~15 min on a fresh VM.

chemistry-sourabh:
why is it creating another VM ?

sirushtim:
@chemistry-sourabh The goal of the developer script is to provision a VM from BMI instead of a physical node. So `bmi pro` is going to setup the environment and when we start the VM, it will boot off the configuration files BMI sets up for us. This will act as an integration test for us.

chemistry-sourabh:
I believe that will be a bad test as there is no guarantee that the bare metal node will boot if the VM does. (Cloud images for example)
We can stick to using a mock node in HIL and do the actual booting thing manually or use Openshift to automate it.
Probaby @ravisantoshgudimetla will agree ?

sirushtim:
@chemistry-sourabh We can't expect every developer to have baremetal node access, the necessary networks setup to the hardware for obvious reasons. We can have a separate CI that will have the necessary networks and hardware set up. That's the real motivation for deploying to the Virtual Machine.

> I believe that will be a bad test as there is no guarantee that the bare metal node will boot if the VM does. (Cloud images for example)

Well, I believe no test that appropriately tests any part of the code is a bad test :-), but instead, we should think of it as a good sanity test the developer can run locally before they push the code. With the different variants of hardware we have out there, there will be a CI setup for each type of node that BMI intends to support and that will actually test if the node will boot or not. Since in the long term, we can't expect every developer to have every type of hardware and expect them to check if their code works on every type of hardware as that is counter productive.

chemistry-sourabh:
I believe a HIL mock node will be enough as we will be doing the actual boot tests. All Developer may not have that much memory on their systems. The dev environment should be as light weight as possible. If it works with a mock HIL node it should work with all.

Also this test doesnt test any new code which the mock node  will not :)

sirushtim:
@chemistry-sourabh AFAICT, a mock HIL node is just a passthrough in the code. Perhaps, you have another idea in mind on how we should be testing these with passthrough code. 

Here are some scenarios I have in mind behind my motivation for a virtual machine which if you think can be solved without it, I'd be happy to not create a virtual machine:

- How do we test if the iSCSI target we've created actually works? We can verify that a target is created but not know if it is actually functional.
- How do we know if the image from Ceph is being booted by a node?
- How do we know if the ipxe template in the code actually works?
....

chemistry-sourabh:
All 3 are addressed by the integration tests which we will run using openshift on our end.
For 1. and 2. if the target was created successfully we can assume they work ( If ISCSI and Ceph dont have bugs, which is not the dev's headache ) 
3. we can write a test case that checks the file being generated is correct. Above that it might work on the VM, but not the bare metal machine as it depends on lot of stuff like lkrn, etc.
For 1. there might be some iscsi test cases we can use.

So my point is it is pointless do so as we have to run most of the integration tests on the bare metal nodes. We do Bare metal provisioning not VMs.

sirushtim:
> All 3 are addressed by the integration tests which we will run using openshift on our end.

FWIW, this Virtual Machine will be used as a base integration test that will be run by a developer before submitting the code. It's often unclear the changes a developer breaks something functionally for obvious reasons. This VM shows visual proof that it works.

> For 1. and 2. if the target was created successfully we can assume they work ( If ISCSI and Ceph dont have bugs, which is not the dev's headache )

We definitely don't want to be ignorant by assuming they just work, we want to test it for ourselves and see if they work and the point is we won't have hardware all the time to see if the "integration" of all the services works.

> 3. we can write a test case that checks the file being generated is correct. Above that it might work on the VM, but not the bare metal machine as it depends on lot of stuff like lkrn, etc.

I think you're misunderstanding semantic checking of files for functional testing. The base case that works for ipxe sanboot always works for virtual machines and hardware. Special hardware needs special tweaking which is obvious.

> For 1. there might be some iscsi test cases we can use.

You're proving my point here. There may be some test cases, but that requires more time to write and BMI already doesn't really have any testing in place which is what we want to address. So we have an integration test to validate the end goal of the command.

> So my point is it is pointless do so as we have to run most of the integration tests on the bare metal nodes. We do Bare metal provisioning not VMs.

Look, the idea is to get developers get visual confirmation that the code they write works and that there is an end result they can visually see. Bluntly assuming they will know what iSCSI targets to check and what ceph image was created and every other process in the chain after they execute a command in BMI is asking too much from them. That is why we want to give them confidence before submitting a new PR that all of it when integrated works through a virtual machine. We need to make sure it is super comfortable and quick for new developers to hack on and submit PR's. User/operator/developer experience should always be prioritized than coming up with ideal scenarios.

I would really like more feedback on solving the issue of BMI throwing an error under this circumstance than digressing and discussing about which developer methodology to use so we can make some progress on this PR.

chemistry-sourabh:
Comment it out for now. Did you try booting a VM from BMI ? I never tried it.

chemistry-sourabh:
Also Register a Mock Node. I would prefer to make sure my integration tests are passing on the mock setup before moving to the actual thing!!

chemistry-sourabh:
Another thing which I didn't like, the script dumps everything related to dependencies where you run from. Like hil folder, ceph related files, etc. Its better to clone HIL in home and dump the ceph files in /etc/ceph/ for clarity

chemistry-sourabh:
I also got the error which @naved001 faced. Based on my rough analysis, there is no tgtd service which the tgt driver trys to start. So @sirushtim I have no clue on installing TGT @apoorvemohan is your guy

chemistry-sourabh:
Things I faced issues with.
Forgot to remove the extra ims clone which caused issues
No HAAS_USERNAME and HAAS_PASSWORD set so caused issues with CLI commands. Should add them to .bashrc also.
For some reason the sqlite3 insert raised a no table called project. Ran it manually worked!!
Better do a bmi mv after the import to change the name from cirros... to test_image or can wget with the required name to avoid delays.

naved001:
@sirushtim Now that the config PR has been merged, you might want to update the sample cfg file that your script loads.

chemistry-sourabh:
@naved001 add it to the list on top

chemistry-sourabh:
@sirushtim I think you better update your script with required changes. Also rebase this branch.

sirushtim:
@chemistry-sourabh 

> Comment it out for now. Did you try booting a VM from BMI ? I never tried it.

It does.

> Also Register a Mock Node. I would prefer to make sure my integration tests are passing on the mock setup before moving to the actual thing!!

That's already being taken care of in the script. +haas node_register bmi_node mock moch-hostname mock-username mock-password? Do try reading the script again.

> I also got the error which @naved001 faced. Based on my rough analysis, there is no tgtd service which the tgt driver trys to start. So @sirushtim I have no clue on installing TGT @apoorvemohan is your guy

tgtd is installed. As discussed, we're missing a specific tgt variant for ceph. I'm trying that out.

> Things I faced issues with.
> Forgot to remove the extra ims clone which caused issues
> No HAAS_USERNAME and HAAS_PASSWORD set so caused issues with CLI commands. Should add them to .bashrc also.
> For some reason the sqlite3 insert raised a no table called project. Ran it manually worked!!
> Better do a bmi mv after the import to change the name from cirros... to test_image or can wget with the required name to avoid delays.

.bashrc is not something we should be writing to. We should just have a script to source once the script is over and add it to README. sqlite3 is not something that I could reproduce. Is the name change from cirros to test_image effectively going to change anything, it doesn't make sense.

@sirushtim I think you better update your script with required changes. Also rebase this branch.

Yes, it seems like it's time to update this script now that we know what to do with TGT.

chemistry-sourabh:
Changing the name is for convenience when running CLI. The dev will need to type the entire long name and might get it wrong also (happened to me)

sirushtim:
@chemistry-sourabh The answer to that will be bash completion and not changing the name. `test-image` is just going to mislead people using it making them wonder what image type.

chemistry-sourabh:
not test-image, atleast cirros-test-image or something that is shorter and easier. Bash completion sounds like a great idea! will raise issue for it.

sirushtim:
Still more commits to come...

naved001:
@chemistry-sourabh I agree, let's call it `cirros-test-image`. we can also write about it on the developer documents that "hey we are using cirros-test-image just to test it". As much as I like bash completion, we should not count on it right now. Let's have bash completion once we have a more stable version.

chemistry-sourabh:
@sirushtim this still doesn't install TGT correctly right ? I remember @apoorvemohan telling that TGT doesn't work on ubuntu, you will probably need to change it to Centos

sirushtim:
Do not merge yet. Needs some more testing.

chemistry-sourabh:
@sirushtim is this ready for review ?

apoorvemohan:
@sourabh: yes

chemistry-sourabh:
Ok will review asap

ims/einstein/ceph.py
```diff
@@ -48,7 +48,7 @@ def __init_context(self):
 
     @trace
     def __init_cluster(self):
-        cluster = rados.Rados(rados_id=self.rid, conffile=self.r_conf)
+        cluster = rados.Rados(conffile=self.r_conf)
```
chemistry-sourabh: Did anybody check this by running as non root user ?

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,64 @@
...
+
+# Move to main ims directory for install from script location
+pushd $TOP_DIR/../../
+
+### Install BMI
+
+user=`whoami`
+sudo mkdir -p /var/log/bmi/logs /etc/bmi /opt/bmi
+sudo chown -R $user:$user /etc/bmi /var/log/bmi /opt/bmi
+sudo chmod 775 /etc/bmi /var/log/bmi /opt/bmi
```
chemistry-sourabh: Why can't we run the entire script as root instead of running each command as sudo ?

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,64 @@
...
+
+### Copy config/template files
+cp bmi_config.cfg.test /etc/bmi/bmiconfig.cfg
+cp ims/ipxe.temp /etc/bmi/ipxe.config
+cp ims/mac.temp /etc/bmi/pxelinux.cfg
+
+virtualenv .bmi_venv
+source .bmi_venv/bin/activate
+pip install python-cephlibs
+python setup.py develop
```
chemistry-sourabh: difference between develop and install ?

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,64 @@
...
+# This is weird
+sudo chmod 777 /etc/tgt/conf.d/
+
+popd
+
+bmi db ls
+
+### Upload Image to Ceph
+wget https://launchpad.net/cirros/trunk/0.3.0/+download/cirros-0.3.0-x86_64-disk.img
+rbd import cirros-0.3.0-x86_64-disk.img --dest-pool rbd
```
chemistry-sourabh: Cant pool be given through a variable ?

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,64 @@
...
+
+### Interface image to BMI
+bmi import bmi_infra cirros-0.3.0-x86_64-disk.img
+
+cat >bmi_userrc.sh <<EOL
+export HAAS_ENDPOINT=http://127.0.0.1:5000
+export HAAS_USERNAME=haas
+export HAAS_PASSWORD=secret
```
chemistry-sourabh: HAAS_ENDPOINT is not required for bmi.

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,64 @@
...
+cp ims/mac.temp /etc/bmi/pxelinux.cfg
+
+virtualenv .bmi_venv
+source .bmi_venv/bin/activate
+pip install python-cephlibs
+python setup.py develop
+
+touch /opt/bmi/bmi.db
+
+export HAAS_ENDPOINT=http://127.0.0.1:5000
```
chemistry-sourabh: No need to export haas_endpoint here

scripts/install/install_hil.sh
```diff
@@ -0,0 +1,59 @@
...
+# Create Haas projects
+haas project_create bmi_infra
+
+### Setup HaaS mock node
+haas node_register bmi_node mock moch-hostname mock-username mock-password
+haas project_connect_node bmi_infra bmi_node
+
+### Setup Mock Node that BMI deploys to
+qemu-img create -f qcow2 /opt/bmi_virtual_node.qcow2 8G
+sudo virt-install -n BMIVM  --description "Test BMI VM" --os-type=Linux --os-variant=rhel6 --ram=2048 --vcpus=2 --disk path=/opt/bmi_virtual_node.qcow2,bus=virtio,size=10 --pxe --network bridge:virbr0 &
```
chemistry-sourabh: I thought we agreed to make this optional ?

scripts/install/install_packages.sh
```diff
@@ -0,0 +1,74 @@
...
+
+die() { echo "$@" 1>&2 ; exit 1; }
+
+# Determine Vendor
+
+OS=$(lsb_release -i -s)
+
+if [[ $OS =~ "Ubuntu" ]]; then
+    sudo apt-get update
+    sudo apt-get upgrade -y
```
chemistry-sourabh: This should be removed as discussed

scripts/install/install_bmi.sh
sirushtim: We could except then everything we install will be root owned and then possibly, we'd regress on permission issues. We need to explicitly define what is ours and own them.

scripts/install/install_bmi.sh
sirushtim: http://lmgtfy.com/?q=difference+between+setup.py+install+and+develop :P

scripts/install/install_bmi.sh
sirushtim: Why?

scripts/install/install_bmi.sh
sirushtim: This is left mostly just for consistency, I guess we could remove it.

scripts/install/install_hil.sh
sirushtim: Yeah, this would be a nice to have feature but I think it needs more work. We shall remove it and add it later.

scripts/install/install_packages.sh
sirushtim: Yeah, this sounds good.

scripts/install/install_bmi.sh
chemistry-sourabh: Realized this is specific to CI or dev, but what if we use it to deploy ?

scripts/install/install_bmi.sh
apoorvemohan: For this release, BMI will be deployed-by and running-as the "root" user.

ims/einstein/ceph.py
apoorvemohan: @naved001

scripts/install/install_bmi.sh
chemistry-sourabh: I need to give a second look at the permissions. But we figured out the permissions.

apoorvemohan:
@chemistry-sourabh @naved001 are we good to merge this?

chemistry-sourabh:
@apoorvemohan suggested some changes if I remember right ? Will give a second look at the permissions tonight.

naved001:
@apoorvemohan I ran this on RHEL and it worked, but it didnt work on ubuntu. We discussed that on github. I actually don't remember exactly what the problem was but I remember sharing the logs with @sirushtim.

scripts/install/install_bmi.sh
chemistry-sourabh: We don't need to chmod here, just chowning is sufficient.

Also /var/lib/tftpboot and /var/lib/tftpboot/pxelinux.cfg also need to be chowned.

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,62 @@
...
+# Temporary changes to get stable CI. Need to do pull request to get this fixed.
+sed -i s/"HAAS_BMI_CHANNEL = \"vlan\/native\""/"HAAS_BMI_CHANNEL = \"null\""/ ims/common/constants.py
+
+python scripts/einstein_server &
+python scripts/picasso_server &
+
+sleep 5
+
+# This is weird
+sudo chmod 777 /etc/tgt/conf.d/
```
chemistry-sourabh: chowing this sufficient no need to chmod it. Tested also if anybody is wondering.

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,62 @@
...
+
+user=`whoami`
+sudo mkdir -p /var/log/bmi/logs /etc/bmi /opt/bmi
+sudo chown -R $user:$user /etc/bmi /var/log/bmi /opt/bmi
+sudo chmod 775 /etc/bmi /var/log/bmi /opt/bmi
+
+### Copy config/template files
+cp bmi_config.cfg.test /etc/bmi/bmiconfig.cfg
+cp ims/ipxe.temp /etc/bmi/ipxe.config
+cp ims/mac.temp /etc/bmi/pxelinux.cfg
```
chemistry-sourabh: Although this is pointless now as BMI won't use these, but stick to cfg/config

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,62 @@
...
+source .bmi_venv/bin/activate
+pip install python-cephlibs
+python setup.py develop
+
+touch /opt/bmi/bmi.db
+
+export HAAS_USERNAME=haas
+export HAAS_PASSWORD=secret
+
+bmi db ls
```
chemistry-sourabh: I think I forgot this, why do need a touch again when we are doing db ls ?

scripts/install/install_bmi.sh
chemistry-sourabh: Also better do this tgt thing before running BMI.

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,62 @@
...
+python scripts/picasso_server &
+
+sleep 5
+
+# This is weird
+sudo chmod 777 /etc/tgt/conf.d/
+
+popd
+
+bmi db ls
```
chemistry-sourabh: This is to check if it is working right ? Do a bmi ls as it will confirm picasso is also setup properly.

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,62 @@
...
+
+### Upload Image to Ceph
+wget https://launchpad.net/cirros/trunk/0.3.0/+download/cirros-0.3.0-x86_64-disk.img
+rbd import cirros-0.3.0-x86_64-disk.img --dest-pool rbd
+
+### Interface image to BMI
+bmi import bmi_infra cirros-0.3.0-x86_64-disk.img
```
chemistry-sourabh: Can this import thing be made into an if else thing like if is for CI/Dev do the import or else don't ?

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,62 @@
...
+
+cat >bmi_userrc.sh <<EOL
+export HAAS_USERNAME=haas
+export HAAS_PASSWORD=secret
+EOL
+
```
chemistry-sourabh: Can we kill einstein and picasso at the end ?

scripts/install/install_bmi.sh
sirushtim: BMI doesn't use these template files? and also what cfg/config?

scripts/install/install_bmi.sh
sirushtim: hmm, we should probably do another iteration if we're serious about running this for non-ci/dev changes. There'd be more changes needed though.

scripts/install/install_bmi.sh
sirushtim: The user is probably going to expect BMI to be running after s(he) installs the script.

scripts/install/install_bmi.sh
sirushtim: Yeah, that sounds good.

sirushtim:
@naved001 I think there were some iterations after that to fix it on Ubuntu because of a libvirt networking conflict, please do try again and let me know if you still face the issue on Ubuntu.

scripts/install/install_bmi.sh
chemistry-sourabh: Ok why will the user expect BMI to be running at the end ? This is an installation script, I have never seen an installation script which runs the program without atleast giving a checkbox or question at the end.

scripts/install/install_bmi.sh
chemistry-sourabh: My only concern is this will also be the official deployment script for 0.5. So people wont want to import cirros into ceph.

scripts/install/install_bmi.sh
chemistry-sourabh: I meant BMI doesn't pick up the templated from /etc/bmi yet. ipxe.config/cfg pxelinux.cfg/config .

naved001:
@sirushtim have you tried running it?

sirushtim:
@naved001 Yes, it worked for me.

scripts/install/install_bmi.sh
sirushtim: Considering, it's `0.5` and it's a pre-release, I can only think people will want to have an image ready to test.

scripts/install/install_bmi.sh
sirushtim: Lets say this is more of a deployment script but for developers. Most infra services that offer a script like this generally have the service running in the end. Not a big deal really.

sirushtim:
@naved001 Did you get a chance to test this?

chemistry-sourabh:
@sirushtim I dont know if you know, but there is a way to get vlan/native by changing the HIL config.
@naved001 knows how to do this. So maybe that line where you modify the constants can be removed.

naved001:
@sirushtim running it now. will keep you posted on this.

ims/einstein/ceph.py
naved001: what does it do? and why are we changing this in this PR?

sirushtim: BMI wouldn't start up without this change strangely for me. So I had to make this change. Can you check if einstein and picasso start up without this change? 

FWIW, it also seems redundant since `rados_id` already exists in the conf.

scripts/install/install_ceph.sh
```diff
@@ -0,0 +1,52 @@
...
+
+ceph-deploy rgw create $node_name
+ceph-deploy mds create $node_name
+
+
+### TEST
+
+# sudo pkill -f ceph
+# sudo losetup -d /dev/loop2 /dev/loop1 /dev/loop0
+# sudo rm /etc/ceph/ceph.conf
```
naved001: you can remove the commented out code.

ims/einstein/ceph.py
chemistry-sourabh: I didn't have issues in my docker container or before, @sirushtim could you reproduce this ?

scripts/install/install_bmi.sh
chemistry-sourabh: @sirushtim check this out

sirushtim: Sounds good.

naved001:
@sirushtim can you address @chemistry-sourabh's comments on this PR.

sirushtim:
@naved001 Yes, on it!

ims/einstein/ceph.py
sirushtim: This is the error I get if I don't make this change:

https://pastebin.com/kMsE4ij9

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,58 @@
...
+TOP_DIR=$(cd $(dirname "$0") && pwd)
+
+# Move to main ims directory for install from script location
+pushd $TOP_DIR/../../
+
+### Install BMI
+
+user=`whoami`
+sudo mkdir -p /var/log/bmi/logs /etc/bmi /opt/bmi /var/lib/tftpboot
+sudo chown -R $user:$user /etc/bmi /var/log/bmi /opt/bmi /var/lib/tftpboot /etc/tgt
```
chemistry-sourabh: this should be /var/log/bmi/logs right ? or the other way around ?

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,58 @@
...
+
+### Copy config/template files
+cp bmi_config.cfg.test /etc/bmi/bmiconfig.cfg
+cp ims/ipxe.temp /etc/bmi/ipxe.cfg
+cp ims/mac.temp /etc/bmi/pxelinux.cfg
+
+virtualenv .bmi_venv
+source .bmi_venv/bin/activate
+pip install python-cephlibs
```
chemistry-sourabh: No need to install this anymore since ceph-common is going to do it

scripts/install/install_bmi.sh
```diff
@@ -0,0 +1,58 @@
...
+touch /opt/bmi/bmi.db
+
+export HAAS_USERNAME=haas
+export HAAS_PASSWORD=secret
+
+bmi db ls
+sqlite3 /opt/bmi/bmi.db "insert into project values (0, 'bmi_infra', 'bmi_network')"
+
+# Temporary changes to get stable CI. Need to do pull request to get this fixed.
+sed -i s/"HAAS_BMI_CHANNEL = \"vlan\/native\""/"HAAS_BMI_CHANNEL = \"null\""/ ims/common/constants.py
```
chemistry-sourabh: What did we decide on this ?

sirushtim: Let's come back to this with a revision!

scripts/install/install_bmi.sh
sirushtim: Since we're running bmi in a virtual environment, we need to install python-cephlibs for rados client in bmi to work.

Also, ``apt-cache showpkg ceph-common | grep python-cephlibs`` doesn't yield anything

scripts/install/install_bmi.sh
chemistry-sourabh: My only concern is people might accidently commit it. Since we have an alternative

scripts/install/install_bmi.sh
sirushtim: Could you elaborate on your question?

scripts/install/install_bmi.sh
chemistry-sourabh: ok then it makes sense. ceph-common installs python-rbd and python-rados which have the cephlibs (Tested this). Since we are using virtualenv we will need it as it uses separate site-packages

scripts/install/install_bmi.sh
chemistry-sourabh: You made a directory called /var/log/bmi/logs as root and configured BMI to dump logs in it.
You are chowning /var/log/bmi which will not allow BMI to write to the /logs subfolder as it will be owned by root and BMI will run as $user.

chemistry-sourabh: Oh the -R option my bad

pgrosu:
Just out of curiosity, why not let Travis run this PR as a test install?

chemistry-sourabh:
The builds will be slow as it will have to install everything every time. HIL dependencies take some time to install. No idea how long ceph dependencies will take.

pgrosu:
Understood.  Based on the scripts, I should be able to run a mock HIL and BMI on a vm via this command:

`cd scripts/install/ ./install.sh`

Is this correct, or am I missing something?  Also what Linux distribution and version do you recommend I could try out this integration test?  It might be nice to have something like Jenkins eventually.

Thanks,
Paul

pgrosu:
Update from BMI meeting:

Ubuntu 16.04

pgrosu:
Which VM is this test running on?

djfinn14:
I tried running the install script  from a clean Ubuntu 16.04 VM on the MOC cluster and it just hung at line 192 of ceph.py for about 10 minutes before I stopped it. Any ideas what I am missing?

image

sirushtim:
@djfinn14 Looks like a permissions issue for Ceph from the latest patchset, I'll update the script in a bit. Thanks.

djfinn14:
Thanks @sirushtim , I also noticed that if i just run the scripts individually, after running install_packages and then install_ceph when I check the status of ceph it produces a warning. It seemed to be that the issue was  with the weights associated with the OSDs. They all had a weight of 0. I ran the command "ceph osd crush reweight osd.0 0.001" for all 3 OSDs and shortly thereafter the status turned into HEALTH_OK. Not sure if this is helpful at all or if it is a nonfactor but figured I would let you know.

sirushtim:
So based on some experiments, it seems like the change in the osd journal size to 1024 MB makes ceph hang at times. I've reverted to the original 10000MB limit and it's working fine again. We'll probably have to find the right way to reduce the journal size later. It's unlikely well need bigger disks though as the script creates sparse files and won't really occupy more disk space but this is untested.

@djfinn14 Can you pull the latest changes and see if it works for you?

djfinn14:
@sirushtim it is working better now. I got a "success" message when provisioning the bmi_node. However there is still an intermittent problem when running install_ceph.sh. There is a race condition that is occurring between the creation of the 
image and the gathering of that same keyring that results in the script erroring out. Paul and I discovered this and I was able to reproduce it today. We know it is a timing issue because we looked at the timestamp of the error and compared it to the timestamp of the keyring file creation.

djfinn14:
Here is the output and errors from bmi db ls
[error.txt](https://github.com/CCI-MOC/ims/files/1065037/error.txt)
[output.txt](https://github.com/CCI-MOC/ims/files/1065039/output.txt)

sirushtim:
@djfinn14 The logs seem fine, there aren't ERROR labeled messages.

Do you have additional data as to where ceph.gather_keys fails? So if the keyring can't be created by the time ceph.gather_keys is invoked, we'd have to add say `sleep 10` if possible to the ceph-deploy command sequence or get a newer version of the ceph-deploy package from their github repository should they have this bug fixed.

FWIW, I ran the script twice and haven't run into this issue yet. I'll keep trying and let you know if I happen to run into it.

djfinn14:
I believe it is because sudo apt-get install ceph-deploy installs and older version of ceph-deploy (1.5.32) whereas they are up to 1.5.38. I have been pip installing 1.5.33 and ran it 3 times without the error.

The error occurs from the line ceph-deploy --overwrite-conf mon create-initial I think, and this line both creates and gathers the keys so not sure where you would put the sleep.

Here is the file creation timestamp and the error with timestamp:

image

You can see it tries to gather it at 20:10:17,066 but it isn't created until 20:10:17,086

sirushtim:
@djfinn14 Thanks for the logs. So if a greater version of ceph-deploy seems to solve it, lets update the script to install ceph-deploy via pip than apt-get.

djfinn14:
Just ran through the script a handful of times and was able to provision and deprovision successfully after each iteration. The only thing is still that it displays the debug information in the terminal with every bmi command, not sure if this is intentional or not.

pgrosu:
@djfinn14 That's great!  Almost there.

djfinn14:
Thanks @chemistry-sourabh, changing it to false worked. Is that something that should be set to true by default? The script is installing everything correctly now judging by my repeated installs, this is more of an aesthetic question.

chemistry-sourabh:
@djfinn14 you are right @sirushtim can you please set the verbose to false in the config file?

chemistry-sourabh:
Also @sirushtim did you finish implementing required changes ?

sirushtim:
@djfinn14 @chemistry-sourabh The reason we don't set verbose to false is so that we can have logs in the CI that will help us figure out the solution to a possible bug in the code.

@chemistry-sourabh Everything should be fixed except for changing the constant to null as that would need more config changes. I would suggest we merge this and then file a bug to fix that separately.

djfinn14:
@sirushtim correct me if I am wrong, but won't we still log the information in a file as long as debug is set to true? My comment was that it is showing up in the terminal which makes it hard to see the actual output of a given command.

djfinn14:
The CentOS install should also use pip to install ceph-deploy to make sure you get the newest version, yum installs a much older one which causes errors.

sirushtim:
@djfinn14 Oh yeah, that works too, basically, I wanted the verbose output in the CI runs. If we have a separate log file, that works too. Thanks, I'll set it to false.

Sure, the RHEL version will also use pip in a following patch.

pgrosu:
Not quite working yet.  I tested it and got some errors.  Here is gist link of the test result:

https://gist.github.com/pgrosu/79410b1c4fbbdaa930f67d2a6f211719

Thanks,
Paul

sirushtim:
@pgrosu This was tested on RHEL 7 and never on CentOS.

pgrosu:
Could we run this test again after a rebase as we feel this code seems to based on an older dev version.

Thanks,
Paul

pgrosu:
We were wondering if you could please rebase this PR so that we ensure it passes the most-up-to-date dev branch.

Thanks,
Paul

naved001:
The HIL team changed the name from HaaS to HIL, which breaks this. So we can either git clone an older commit, or change the name to HIL here (I'd prefer the former). @sirushtim would you have time to work on this issue? Let's just make one final patch to take care of the most important issues and merge this because it has been sitting for too long.

sirushtim:
Sure, I'll try rebasing this and will post a revision soon.

chemistry-sourabh:
Also @naved001 why not merge Kan's HIL PR ?

naved001:
Important note: the ceph that this script sets up fails if we reboot the machine. We have to make the loopback devices again, re-read the partition table (`partprobe /dev/loopx`) and then "ceph activate all osds".

sirushtim:
@chemistry-sourabh @naved001 So, are we merging Kan's PR before this?

naved001:
I don't want to deal with the merge conflicts that that PR might introduce. I think we should close that PR for now and redo that once we have no other outstanding PRs.

chemistry-sourabh:
Those conflicts will happen even if you have no outstanding PRs (which I highly doubt will happen). I suggest merging for the 0.5 release as we will want all haas's to hil's.

@naved001 try it out, if it is too tough I can do it in the evening.

chemistry-sourabh:
@sirushtim there are still a lot of unanswered stuff in this thread. Please make a separate issue regarding them. Also make sure this works before merging :P

naved001:
@apoorvemohan I am removing the 0.5 milestone because I think this PR, though really helpful, is not core BMI and does not affect the main codebase or the upcoming redesign.

apoorvemohan:
@naved001 I don't get it - do you want to delete the 0.5-release completely or just remove this PR from the 0.5-release?

chemistry-sourabh:
Lol @apoorvemohan I think @naved001 did a typo. I think he means to remove this PR.

naved001:
yeah, what @chemistry-sourabh said,. just removing this PR from the milestone. If @sirushtim addresses the comments on this PR and it works then Ill just merge it.

sirushtim:
Haven't really had the time to test this yet, if someone can, please do so! 

As @chemistry-sourabh suggests, it'd be better to track the unanswered issue of the config change separately!

apoorvemohan:
I don't agree to removing this PR from 0.5-release. We have to provide a stable (though limited) way to setup BMI in this release. How do you expect people using this PR to setup BMI in their environment without a stable release script?

naved001:
> I don't agree to removing this PR from 0.5-release. We have to provide a stable (though limited) way to setup BMI in this release. How do you expect people using this PR to setup BMI in their environment without a stable release script?

I am convinced. Putting back the milestone on it.

djfinn14:
Just tested the new changes by @naved001 several times, the scripts work and I was able to provision and deprovision on Ubuntu, but not on CentOS

chemistry-sourabh:
Damn @naved001 now I will have to review this again :(

naved001:
@chemistry-sourabh @sirushtim @pgrosu 

@djfinn14 got this working on centos too with the  new changes. Now it works for both. I need one more plus one because Dan and I collaborated on this.

naved001:
Sure. We aren't guaranteeing RedHat with this script though. It works on
CentOS and Ubuntu, and may work on RedHat.

pgrosu:
I'm still running the UAT but they are very slow on Kumo...

bmi_config.cfg.test
```diff
@@ -0,0 +1,53 @@
...
+ipxe_path = /etc/bmi/ipxe.cfg
+
+# this section is for http config
+[rest_api]
+ip = 127.0.0.1
+port = 9000
+
+# this section is for logs
+[logs]
+path = /var/log/bmi/logs
```
pgrosu: Please change this line to the following - by appending a forward slash (`/`) at the end:
```
path = /var/log/bmi/logs/
```
It's causing logs to be written to the incorrect directory, as follows - causing UAT to fail:

```
$ ls -ltr /var/log/bmi/logs
total 0
$ ls -ltr /var/log/bmi/
total 280
drwxr-xr-x 2 ubuntu ubuntu   4096 Aug 11 16:56 logs
-rw-rw-r-- 1 ubuntu ubuntu 282601 Aug 14 05:57 logsims.log
$
```
Thanks,
`p

naved001: Good catch! Thanks for pointing this out. Is there anything else? We can push those changes in the next commit.

pgrosu: Thanks, but when would be next commit?  Don't we want this to be validated for the 0.5 release?  Basically I'm running through the UAT to ensure the PR passes for an end-to-end scenario.

naved001:
I'd push the next commit as soon as you point out what changes need to be
made. If that change in path is the only thing, I can do that now.

naved001:
@pgrosu
> Please change this line to the following - by appending a forward slash (/) at the end

Done

naved001:
@pgrosu ping.

pgrosu:
@naved001 I was getting different issues on Friday that I was working through, but now I'm rebuilding the VM to ensure I have a fresh instance and trying the tests again.

pgrosu:
@naved001 Does not look like it is properly working on Centos 7, as `rbd` does not seem to actually get installed properly by the scripts:

```
HTTP request sent, awaiting response... 200 OK
Length: 9761280 (9.3M) [text/plain]
Saving to: ‘cirros-0.3.0-x86_64-disk.img’

100%[======================================>] 9,761,280   10.4MB/s   in 0.9s

2017-08-21 22:20:14 (10.4 MB/s) - ‘cirros-0.3.0-x86_64-disk.img’ saved [9761280/9761280]

+ rbd import cirros-0.3.0-x86_64-disk.img --dest-pool rbd
./install_bmi.sh: line 52: rbd: command not found
```

naved001:
You need to make some changes manually to get rbd to work in those
openstack VMs. Something to do with hostnames(there's a comment in the
script).  We can't make the script make those change because it's specific
to these Openstack VMs.

pgrosu:
@naved001 Let's work tomorrow together on this so that I can incorporate them into the UAT framework for them become automatically triggered and configured for the PR-60 test scenarios.

naved001:
I am sorry but I can't sit together and hack on this. I have already
explained the reasoning behind why it failed CentOS and why we left it that
way. If you have a recommendation on how to fix it (or any other concern)
 then let us know in the comments and we'll push another commit to fix it.

Moreover, you still haven't published your UAT code and it hasn't even been
reviewed by anyone yet; so I don't want to send my time hacking that right
now. And I don't understand how is that linked with this PR anyway. This
setus up BMI, your tests run *after* the setup.

bmi_config.cfg.test
chemistry-sourabh: Is there a specific reason why we are having a 'logs' folder in /var/log/bmi ? I dont see the reason for it.

pgrosu:
Hi Everyone,

Now this PR passes both for Ubuntu 16.04 and CentOS 7.  The results are the following link:

https://gist.github.com/pgrosu/c05e428b5e95fcd81a7a9b1b4c74f149

A few things that were required when testing this on OpenStack VMs:

###### For CentOS

* The `/etc/hostname` would look like this `pgrosu-ubuntu16.04.moclocal`, and it would need to be truncated on the first period to `pgrosu-ubuntu16` for RBD to be installed properly, and find the proper hostname.

* Git would need to be installed as follows: `sudo yum -y install git`

###### For Ubuntu

* Python would need to be installed as follows:

```
sudo add-apt-repository -y ppa:fkrull/deadsnakes
sudo apt-get -y update
sudo apt-get -y install python2.7
sudo ln -s /usr/bin/python2.7 /usr/bin/python
```

###### For either version

* As @djfinn14 noticed, the IP address of the VM would need to be added to the `ceph.cfg` file under the `ims/scripts/install/` directory.  I added for the PR-60 configuration for the UAT the following that automates it:

```
if [ ! -z "`ifconfig | grep inet | head -n1 | grep :`" ]; then
  IP_ADDRESS=`ifconfig | grep inet | head -n1 | cut -f2 -d':' | cut -f1 -d' '`
else
  IP_ADDRESS=`ifconfig | grep inet | head -n1 | cut -f2 -d'i' | cut -f2 -d' '`
fi

echo -e "public_network = $IP_ADDRESS/24\nosd pool default size = 2\nosd crush chooseleaf type = 0" >> ceph.conf
``` 

Nota bene:  I _**highly**_ recommend that for every release of HIL, we get alerted and run this UAT scenario.  Ideally we have a Jenkins CI Server so that it gets triggered by either team as both projects have a connection to each other.

This is now ready to be merged - making the 0.5 release official - and I recommend the `dev` branch to also be merged into `master`.  I will work now on creating a manual for UAT so that folks can use it as well.   

I will put together a set of release notes too.

Thank you everyone for the invaluable help and patience!
Paul

bmi_config.cfg.test
pgrosu: I guess it is redundant.  We can always update it anytime :)

apoorvemohan:
@pgrosu Where can I find the UAT?

Issue #33, opened 2017-01-10
Filesystem Driver


chemistry-sourabh:
Issue #30 should be done before coming to this.
Since the framework will be written just an interface needs to be written, Ceph needs to extend that interface and the framework needs to be used to load the class.

apoorvemohan:
@chemistry-sourabh can we close this?

PR #70, opened 2017-02-18
Partial directory reorganization to support interfaces.

ravisantoshgudimetla:
This is precursor to #33. We need this folder re-organization to make sure we can write mock's for everything and this looks a bit cleaner.


chemistry-sourabh:
I am fine with this structure, but it would be better to put it each driver in its own package like.
einstein/iscsi/tgt/tgt.py
Because my later PRs will be moving the driver related constants from commons/constants to the drivers_package/constants
Also It would be better to call filesystem as fs (for consistency)
Also change networkisolator to network_isolator

chemistry-sourabh:
Also I had this question for debate.
Is it better to leave the drivers in the einstein package or move to something like drivers/ext in the root directory of project (like HIL) ?

ravisantoshgudimetla:
I didn't understand the need for having tgt folder. Any specific reason? 
> Is it better to leave the drivers in the einstein package or move to something like drivers/ext in the root directory of project (like HIL) ?

I think, einstein and picasso should be up for debate. Somehow, I think we can change them to more technical names - controller and conductor etc. I don't know why HIL used ext folder.

ravisantoshgudimetla:
@chemistry-sourabh , approve it as well.

chemistry-sourabh:
The reason for having a tgt folder is to split files. I put the constants for tgt in a separate file, similarly I put the ceph stuff for IET in a separate file. The splitting of the existing drivers can be debated, but in the future we may need to split files.

chemistry-sourabh:
@ravisantoshgudimetla please move the directories to a root level drivers directory as discussed.

ravisantoshgudimetla:
@chemistry-sourabh I didn't understand what you mean by root level structure. Also, I forgot what we discussed. Please provide the context as well. It will be easier for others as well.

chemistry-sourabh:
ok you wanted something like this.
/
   ims
          einstein
                        iscsi
                                tgt.py
                                mock.py
                        ceph

I didn't like this that much as we can't divide the drivers into sub modules if required. So I wanted something like this


You didn't like this as it has too many folders

So we agreed on this
/
   drivers/ext
                      iscsi
                              tgt
                                   tgt.py/driver.py
                                   constants.py
                              mock
                      ceph
   ims

This sort of gives what everyone wants. Also it is similar to HIL

ravisantoshgudimetla:
@chemistry-sourabh - So want to take out the folders out of einstein and picasso?

chemistry-sourabh:
Only einstein why picasso ? Picasso doesn't have any folders in it.

chemistry-sourabh:
After dealing with test cases it is better to move the drivers to ims/drivers/... for consistency in tests directory structure

ravisantoshgudimetla:
Did we agree upon this? I thought, as of now, we wanted to have einstein and other names..

naved001:
I am fine with this reorganization. @chemistry-sourabh if you are not happy with this yet and want more discussion, I'd suggest you put a red X (as a part of your review) on it so that someone else doesn't merge it until your concerns are addressed.

ravisantoshgudimetla:
@chemistry-sourabh As you know, I will loose the flow easily as I am into few more things. Provide a sample of what folder organization again. I am not ok with sharing constants as well. What if all the constants needed for a particular file are present in the driver file(like tgt.py having its constants in a function or class at the starting.)

chemistry-sourabh:
I already typed this in a previous comment so I am pasting it here. The idea is to move the drivers folders as described below.

ok you wanted something like this.

/ims/einstein/iscsi/tgt.py
/ims/einstein/iscsi/mock.py
/ims/einstein/ceph/...

I didn't like this that much as we can't divide the drivers into sub modules if required. So I wanted something like this

/ims/einstein/iscsi/tgt/tgt.py
/ims/einstein/iscsi/tgt/constants.py
/ims/einstein/iscsi/mock/...
/ims/einstein/ceph/...
You didn't like this as it has too many folders

So we agreed on this
/drivers/iscsi/tgt/tgt.py
/drivers/iscsi/tgt/constants.py
/drivers/iscsi/mock/...
/drivers/ceph/...
/ims/...

This sort of gives what everyone wants. Also it is similar to HIL

chemistry-sourabh:
Also we are not sharing constants, each driver will have its own constants.py file that will have constants related to it, this is required to make each driver truly independent

ravisantoshgudimetla:
>Also I dont like having constants in the file as they generally grow as the file gains features. It is better to have them in a separate file to avoid the headache of moving them later. This will ensure cleaner code.

Thats when we can have separate files, till then I am ok with not having separate files but if everyone wants it, I can make it.

ravisantoshgudimetla:
@CCI-MOC/bmi-breakers  I am closing this PR off as the changes that I need are already made. The only thing that we need is debating if need constants file for every file or not, which I am not going to do in this PR.

PR #72, opened 2017-02-22
Update tests structure

chemistry-sourabh:
This is to update the tests structure as discussed.
@CCI-MOC/bmi-breakers ravi kind of liked this approach. I dont have a huge problem with it, but I preferred suffixing the test files as 'unit' or 'integration' instead of using the folders.

We can debate in  a separate PR whether a test is an unit or integration later, just focus on the directory structure.

The structure is supposed to mimic the structure of the ims folder.

PR #81, opened 2017-02-26
Tests

chemistry-sourabh:
Fixed the previous existing test cases.
Needs #72 and #42  to be merged before


chemistry-sourabh:
Can review the last commit

tests/database/image/unit/test_image.py
```diff
@@ -41,13 +47,16 @@ def tearDown(self):
```
naved001: Commenting here because I can't comment near the lines that aren't a part of the commit. 

Line 29: `if 'image[0] == 'image 3':`; just reversed the order for a more conventional look (variables to the left, and constants on the right)
Moreover, how about we do this
`assert image[0] == 'image 3'` I don't see a need for a for loop here. we can just access the element by it's index. 

that would be cleaner and the same thing for block starting at line 35. 

Line 41 and 42: how about `assert 'image 2' in images` or if you want to use `assertTrue` we could do `assertTrue('image 2' in images)`. 

and then in the tearDown of this test, is deleting the project only sufficient?  do we not have to delete the images first?

tests/database/image/unit/test_image.py
```diff
@@ -1,13 +1,19 @@
 from unittest import TestCase
 
+from ims.common.log import trace
+from ims.database.database import Database
+
 
 class TestInsert(TestCase):
+    """
+    Creates a project and tests inserting images in several combinations
+    """
```
naved001: 1 line doctring?

tests/database/image/unit/test_image.py
```diff
@@ -41,13 +47,16 @@ def tearDown(self):
 
 
 class TestDelete(TestCase):
+    """
+    Inserts image and deletes it
+    """
```
naved001: this can be a 1 line docstring

tests/database/project/unit/test_project.py
```diff
@@ -21,12 +25,15 @@ def tearDown(self):
```
naved001: in line 20: why is projects 2 dimensional?

tests/einstein/operations/unit/test_operations.py
```diff
@@ -1,40 +1,48 @@
...
 from unittest import TestCase
 
+import ims.common.config as config
 import ims.common.constants as constants
 import ims.einstein.ceph as ceph
+from ims.common.log import trace
+from ims.database.database import Database
 from ims.einstein.operations import BMI
 
 _cfg = config.get()
```
naved001: why `_cfg` if we are using it directly everywhere. `cfg` should be fine

tests/einstein/operations/unit/test_operations.py
```diff
@@ -1,40 +1,48 @@
...
+NETWORK = _cfg.tests.network
 
-EXIST_IMG_NAME = "bmi_ci.img"
-NEW_SNAP_NAME = 'test_snap'
-NOT_EXIST_IMG_NAME = "i12"
-NOT_EXIST_SNAP_NAME = "hello"
+EXIST_IMG_NAME = _cfg.tests.exist_img_name
+NEW_SNAP_NAME = _cfg.tests.new_snap_name
+NOT_EXIST_IMG_NAME = _cfg.tests.not_exist_img_name
+NOT_EXIST_SNAP_NAME = _cfg.tests.not_exist_snap_name
```
naved001: Why does this test get the constants like PROJECT, NETWORK etc from the cfg file?

chemistry-sourabh: It doesnt need real setup, It can be mocked also after the mock drivers are done. But you sought of got the idea that we may want to run the tests on a real setup to make sure that setup was done properly.

tests/einstein/operations/unit/test_operations.py
chemistry-sourabh: So that it is not used outside this module, python doesnt use enforce it, but it is a programmer thing. Any class, variable, method, etc with _ is not supposed to be used outside the module or class.

tests/database/image/unit/test_image.py
chemistry-sourabh: will fix this

tests/database/image/unit/test_image.py
chemistry-sourabh: sure

tests/database/project/unit/test_project.py
chemistry-sourabh: Good catch! I missed this and the tests returned successful as 'p' matched

tests/database/image/unit/test_image.py
chemistry-sourabh: 1st one we need a loop as it is a 2D array, need to think about it more.
2nd one need to change it.
3rd there is a cascade rule in db, when we delete project all images under it are also gone in db.

chemistry-sourabh: I just focused on getting the test cases to pass as I know they passed before.

naved001: > 1st one we need a loop as it is a 2D array, need to think about it more.

Why/How is `image[]` a 2d list? even if that's the case you could do this `image[0][0]`. I still don't understand why we would need for loops.
take your own time to think about. 

> 3rd there is a cascade rule in db

That's nice!

tests/einstein/operations/unit/test_operations.py
naved001: So this test would work if we have a real setup or if we have mock drivers. So maybe we should put a comment saying that it is not a unit test and will not work. 
And if we want to keep this the way it is right now, we could put it under `/tests/integration`, you get what I am saying?

naved001:
We have `good_bmi` everywhere? Is there a `bad_bmi` that you plan on adding? 
this reminds me that we should have negative tests too where we make sure that the code throws errors where we expect it. Perhaps, in a separate PR? @apoorvemohan @xukan what are your thoughts on this?

tests/database/image/unit/test_image.py
chemistry-sourabh: It just doesnt return the names of the images it returns more stuff. I advice looking at the method to get better understanding

tests/einstein/operations/unit/test_operations.py
chemistry-sourabh: I guess you didnt read the other PR. I said that we can debate on which test goes where in a separate PR. I suggest making an issue about sorting the test cases or something

chemistry-sourabh:
I wrote a lot more tests in my original tests which are not part of this branch. Those tests need to be added again.
I just ported the tests which tested the main functionalities like provision, deprovision, etc. You and @xukan can write the other ones. Will force you guys to go through the code also.
Yes you are right there was a bad_bmi in the old tests which basically had wrong username and password kind of stuff.

tests/database/project/unit/test_project.py
chemistry-sourabh: Looks like I was mistaken the db function returns a 2D array

chemistry-sourabh:
@naved001 made changes to docstrings and simplified the asserts have a look

chemistry-sourabh:
@CCI-MOC/bmi-breakers can someone else also please review this ?

tests/unit/picasso/test_rest.py
```diff
@@ -7,6 +7,9 @@
 import requests
 
 import ims.common.config as config
+
+config.load()
```
sirushtim: Why aren't we doing this in setUp? Feels odd that this is being done here.

chemistry-sourabh: for trace to work, it needs to be done there.

naved001:
@chemistry-sourabh Could you please re-base this?

naved001:
@knikolla could you review this please?

tests/unit/einstein/test_operations.py
```diff
@@ -74,18 +87,21 @@ def tearDown(self):
 
 
 class TestCreateSnapshot(TestCase):
+    """
+    Provisions an imported image and creates snapshot
```
knikolla: Can you please put ``TestCreateSnapshot``, ``TestListSnapshots``, ``TestRemoveSnapshot`` in one class called  ``TestSnapshots`` with methods ``test_create_snapshot``, ``test_list_snapshots``, ``test_remove_snapshot``. There is a huge amount of repetition in ``setUp``.

tests/unit/einstein/test_operations.py
```diff
@@ -144,27 +163,33 @@ def tearDown(self):
...
+    def runTest(self):
         pass
 
     def tearDown(self):
         pass
 
 
 class TestListImages(TestCase):
+    """
+    Imports an image and calls list image
```
knikolla: Same comments as for snapshots.

tests/unit/picasso/test_rest.py
```diff
@@ -88,22 +104,26 @@ def tearDown(self):
 
 
 class TestCreateSnapshot(TestCase):
+    """
+    Calls provision like TestProvision then creates a snapshot using rest call
```
knikolla: Same comment.

tests/unit/picasso/test_rest.py
```diff
@@ -164,30 +188,38 @@ def tearDown(self):
...
+    def runTest(self):
         pass
 
     def tearDown(self):
         pass
 
 
 class TestListImages(TestCase):
+    """
+    Imports an import image and calls the list images rest call
```
knikolla: Same comment.

chemistry-sourabh:
@knikolla I know there is a lot of repetition in the each class. I did that to be consistent with 1 test case per class rule.

knikolla:
How/why did you come up with that rule?

chemistry-sourabh:
I felt that each class sort of isolates each test case and it will be easy to tweak them without effecting other test cases. If we combine them now and later need to make changes to one of the test cases we may need to split them again.

knikolla:
It's easy to split back. Don't copy paste large amounts of code.

chemistry-sourabh:
ok

chemistry-sourabh:
Looked at the code and found that setup/teardown is varying with a method call or 2. How can I combine them ?

knikolla:
@chemistry-sourabh make it part of the test. I assume the differences you are talking about is the tests for list having the create as part of setup. While the test for create does not.

chemistry-sourabh:
@knikolla I believed that test should have the code which you are testing and everything else goes in setup and teardown. So is it justified to move the create in list to the test as we are not testing it ?

knikolla:
@chemistry-sourabh you can leave the create in setup if you want, and have another create in the test for create, and use what was created in setup for the test for listing.

chemistry-sourabh:
@knikolla have a look

knikolla:
@chemistry-sourabh I don't like how the tests are grouped by operation. Please refer to my prior comment and group the tests by resource type. Like image/snapshot/etc.

chemistry-sourabh:
@knikolla I am confused now. You wanted all the snapshot tests in one class as there was redundancy in setup and teardown.

Do you want to group list_images and remove images also ?

knikolla:
@chemistry-sourabh 

```python
class TestX:
    def test_create_X
    def test_list_X
    def test_delete_X
```

The above is what I had in mind. You can have multiple tests in one class and have them share the setup function. You have done this for snapshots, but not for images. Also why is listing and creating images in test_operations?

chemistry-sourabh:
You mean list and remove image ? They are operations of bmi and are in the operations.py file.

knikolla:
@chemistry-sourabh whereas test_images.py only test db operations?

chemistry-sourabh:
Yes, both test_project and test_images test db operations.

knikolla:
@chemistry-sourabh what is preventing you from putting TestListImages and TestDeleteImages in the same class as different methods? They have the exact same ``setup`` and they touch the same type of resource. 

```python
class TestImageOperations:
    def setup
    def test_list_images
    def test_delete_image
```

chemistry-sourabh:
ok will do it @knikolla

chemistry-sourabh:
@knikolla ok I realized this when trying to merge the images also.

In the snapshots test class the teardown is not deleting the additional snapshot that is being created, which is causing an additional image in ceph to be left after successful running of tests.

There is a similar issue with teardown of list and remove image tests.

Cant we keep the tests in the original way and reduce redundancy by using functions ? I am open to other solutions

knikolla:
@chemistry-sourabh Wait. You are connecting to a real ceph in the unit tests?

chemistry-sourabh:
We dont have a mock driver yet so testing with real ceph.

sirushtim:
Yeah, unit tests aren't supposed to connect to physical resources. Try the unittest.mock framework instead: https://docs.python.org/3/library/unittest.mock.html

knikolla:
@chemistry-sourabh Then they are not unit tests.

naved001:
so how about we change call this PR integration/deployment tests, and save the unit tests for later. Because, obviously, these are testing some stuff (real hardware or not) that we want.

ravisantoshgudimetla:
@chemistry-sourabh If these tests are unit tests, you should not have logic sprawl across components. If you have some then there is something fishy.

chemistry-sourabh:
In my previous PR I said we can decide whether a test is a unit test or integration test later. Now might be good time to do that. If you guys feel this is an integration test, then we can move it to a separate folder.

This just tests the calls in the operations.py file, but it touches ceph, iscsi and hil. I was hoping it will become a proper unit test after the mock drivers are done.

knikolla:
@chemistry-sourabh move them to a different folder for now. I can help with coding the mock drivers once the repo is more stable.

chemistry-sourabh:
@knikolla you want me to move to the integration folder without combining the test cases right ?

knikolla:
@chemistry-sourabh yes.

chemistry-sourabh:
@knikolla Have a look

PR #82, opened 2017-02-26
Test docs

chemistry-sourabh:
Added Testing.md needs #42 and #72 to be merged for better diffs. Can review only the last commit


docs/testing.md
```diff
@@ -0,0 +1,138 @@
...
+       pass
+
+  def tearDown(self):
+       pass
+ ```
+ 
+* Each class's name should start with Test
+* Each setUp method should use the trace decorator
+* setUp method is supposed to do stuff which is required before the actual test
+is run like adding db entries, etc.
```
naved001: can you add what's the usual setup required? it would be helpful if you could list things here
I can think of 
* initialise a new db 
* load the config file
* start einstein and/or picasso

docs/testing.md
```diff
@@ -0,0 +1,138 @@
+# Testing BMI
+
+Every Developer will need to test his/her feature or make sure that he/she
+did not break anything. So the test cases are situated in
```
naved001: "Every developer will need to test **their** feature or make sure that **they** did not break anything" 
Just minor nitpick :) because it sounds better this way.

docs/testing.md
```diff
@@ -0,0 +1,138 @@
...
+
+All of these folders are python packages so they should have a \_\_init\_\_.py
+
+## Writing Tests
+
+Create a file in the correct location of the directory structure and make sure
+the name conforms to the regex 'test_\*.py'
+
+We are using python's unittest as our testing framework for now.
+
```
naved001: just out of curiosity, is there a specific reason to use `unittest` instead of `pytest`. I don't have a strong opinion on this, it's just that I am familiar with `pytest`

docs/testing.md
```diff
@@ -0,0 +1,138 @@
...
+We made a testing framework that collects and runs tests. So follow these steps
+
+* Write a proper bmiconfig.cfg
+* Set the variable BMI_CONFIG
+```
+export BMI_CONFIG=<path>
+```
+* Some tests may require Picasso and Einstein to be running so do that if
+required before proceeding.
```
naved001: If a test requires Picasso and Einstein to be running, then the test should start those under `SetUp`. That way we can run these tests in Travis. What do you say?

docs/testing.md
chemistry-sourabh: Is there anything in pytest we can use ? Why dont you research unittest vs pytest, then we can decide

docs/testing.md
chemistry-sourabh: We cant run it in travis unless we install the entire dev environment. I feel that running picasso and einstein at the beginning of the testcases should be sufficient as we are not storing state in memory. So probably running it in the test framework or run them after installing bmi should be good ?

docs/testing.md
chemistry-sourabh: I didnt add it because it is not a definitive list, it depends on the test.
Like for example the provision and deprovision test cases are actually same, but in provision the deprovision call goes into teardown and provision goes into setup in deprovision test case.

docs/testing.md
chemistry-sourabh: will do this

docs/testing.md
naved001: I am not suggesting the use of pytest. I am fine with unittests, I was only curious.

chemistry-sourabh: I dont know if pytest is better than unittests. So @naved001 please check out the pros and cons of pytest when compared to unittests

chemistry-sourabh:
@naved001 have a look

chemistry-sourabh:
@naved001 updated as per new structure and pytest, please reread

chemistry-sourabh:
@xukan please review this so that we can wrap it up

docs/testing.md
```diff
@@ -101,38 +104,38 @@ Let us know if your test cases need additional variables
...
-be given as regex
-* To run all Test Cases from database and einstein pacakge then database.\* and
- einstein.\* can be given as regex
-* To run all Test Cases .\* can be given as regex
-
-There is an -i option that shows you all the test cases that matched the regex
-before running them.
-
-do python test_ims.py --help for details.
+* Run pytest with the options
```
sirushtim: It would be good to quote the actual command here for beginners to get started.

docs/testing.md
```diff
@@ -101,38 +104,38 @@ Let us know if your test cases need additional variables
 
 ## Running Tests
 
-We made a testing framework that collects and runs tests. So follow these steps
+We have decided to use pytest as out test runner. So follow these steps
```
sirushtim: To be more accurate, you want to mention, this is one of the ways to run tests.

docs/testing.md
```diff
@@ -101,38 +104,38 @@ Let us know if your test cases need additional variables
...
-* To run all Test Cases .\* can be given as regex
-
-There is an -i option that shows you all the test cases that matched the regex
-before running them.
-
-do python test_ims.py --help for details.
+* Run pytest with the options
+* Wait for them to run and see the results.
+
+pytest can take the paths of test files as arguments, this can allow the dev to run
```
sirushtim: Again, some examples would be good.

chemistry-sourabh:
@sirushtim have a look

naved001:
@chemistry-sourabh now that #81 is merged, do we need to make any other changes here or can we merge this too?

chemistry-sourabh:
Better run a sweep to make sure.

chemistry-sourabh:
@naved001 and @sirushtim have a second read asap so that we can merge this.

docs/testing.md
```diff
@@ -0,0 +1,162 @@
...
+* If a test fails, then it is better to refer bmi logs. They will be located in the folder
+specified in config.
+* Delete the old log before running the test again as it may reduce confusion. (You can back 
+it up if you need it).
+* It might be difficult to tell where a particular test started in the logs, just search for
+the test's class name (like TestProvision) in the logs. The first line it hits is the start of
+the test case, then the next hit is the second execution of the same test.
+* --pdb option can be used along with pytest to invoke pdb when error is hit.
+* If getting config errors, double check BMI_CONFIG env variable and the config file.
+
```
naved001: this is slightly unrelated to testing, but since we are talking about debugging in this section; do you want to add a line saying that we should run `rpc_server` and `name_server` separately instead of running `einstein_server`? this took us time to figure out when I was trying to run `pdb`

chemistry-sourabh:
Added @naved001  and @sirushtim

PR #83, opened 2017-02-26
Test framework

chemistry-sourabh:
This is the testing framework. Used click for a better experience. Uses regex find to test cases to run.
Needs #42 and #72 to be merged before. Can Review last commit


tests/test_ims.py
```diff
@@ -1,38 +1,111 @@
...
+def cli():
+    """
+    The Testing Framework for BMI, do test_ims run --help for more details
+    """
+
+
+@cli.command(short_help="Runs the test cases as per the given patterns")
+@click.argument('test_case_patterns', nargs=-1)
+@click.option('-i', '--interactive', is_flag=True, help="Interactive Mode")
+def run(test_case_patterns, interactive):
```
sirushtim: This is basically what `python -m unittest` does which is what developers invoke to run tests generally. This entire file is infact replicating what python -m unittest does. So we should use that or any one of the many testrunners instead. I suggest there be done some research in what fits best and then accordingly remove this file.

chemistry-sourabh: ok I'll have a look

chemistry-sourabh: I am thinking of using pytest @CCI-MOC/bmi-breakers. Is everybody ok with it ?

chemistry-sourabh: But another small issue is that config should be loaded in each test module. Another solution I am thinking is subclassing TestCase and adding config.load() in the __init__(), all test cases will subclass from this.

Another possible solution I can think of is a decorator

sirushtim: Why pytest over unittest? Please include the reasoning.

Subclassing sounds good except we'd need the config loading in setUp() rather than __init__() because then we'd risk race conditions where test cases access the config concurrently since __init__() is called only once per class and setUp() is called for every test case. Any test case wanting to override setUp will just call super and run their own setUp().

chemistry-sourabh: why will there be a race condition ? All the test cases will read the config, not write to it. If we have a test case where something needs to be written to a config then it will be a separate file altogether. I personally dont want to call the super of setup in test cases as it will require modification to all test cases's setup whereas in init just changing the base class is enough. Also the tests are in sequence not parallel.

Why pytest?
1. Couldnt find any documentation related to python -m unittest, what I found states that -m option just runs a lib module and all examples I saw used unittest.main() in test file or discover (which I used in test_ims.py)
2. pytest prints output prettier, you may deny it is unimportant, but it gives more info also.
3. pytest doesnt need any changes to the test cases except the config.load() stuff which will be required by python -m also. (Best part)
4. pytest allows you to attach pdb when error hits. (Also cool)
5. pytest allows to filter by python expressions (like my framework)
6. @naved001 is familiar with pytest :P
There are a lot of options which have to be read carefully.

chemistry-sourabh: I just ran one test module in pytest and it worked. Will run all test cases and see if any issues occur.
Then if everybody is fine will close this PR.

ravisantoshgudimetla:
@chemistry-sourabh , create a patch or when you are committing clone the repo and do a git add of only those files. Its difficult to review when most of files are not needed.

tests/test_ims.py
ravisantoshgudimetla: Ok, I don't have much knowledge on this stuff. I am ok with either of stuff as long as we are not re-inventing the wheels. But let me know, if you want some voting..

chemistry-sourabh:
Just review the last commit

tests/test_ims.py
sirushtim: @chemistry-sourabh 

1. They can be run in parallel.
2. Race conditions will occur when we modify the config files in the future. It's easy to subclass to have a common setUp class. Honestly not that hard. It's almost an anti-pattern to use the __init__ class for test cases. It get's very hard to debug tests that aren't isolated to their own environment.
3. 1-5 is all possible with unittest too, just search and you will find it.
4. Unittest is being used throughout the code.

chemistry-sourabh: @sirushtim  regarding the pytest vs unit test debate, the developer can use either as pytest supports the default unit tests, but not the other way around. There is a pytest way to write test cases that is supposed to reduce boilerplate and make test code smaller, but thats not do that now as we will lose compatability with unit test. Let the developer decide what test runner he wants to use.

Also found this pros and cons https://www.slant.co/versus/9148/9149/~unittest_vs_pytest

But the config loading thing needs to be decided, we can close this PR and move to an issue or continue here.

sirushtim: @chemistry-sourabh setup is called for every test `case` whereas __init__ is called for every Test `Class`.

Like I said, I don't have a strong opinion, do whichever. Just that since there are unittest test cases in the code, we should probably use the unittest test runner instead of the pytest test runner because pytest may not always support all of unittest frameworks test features. We decide one test framework and stick to it.

chemistry-sourabh: I think I found the thing you are talking about.
python -m unittest discover
The -m runs a lib module and discover is a function which scans for test cases. I used discover in the testing framework so the command makes sense, but couldn't find any pdb related flags.
Probably I have to attach pdb old school way.
We will use unittest to write test cases and since unittest and pytest can be used interchangeably we will use pytest for now to run them until we land into some issues with it.
If you are fine with this close this PR. We will continue the discussion of config loading in tests PR.

chemistry-sourabh:
Closing this as it is not required due to shift to pytest

Issue #69, opened 2017-02-18
Unit tests for config object


ravisantoshgudimetla:
@chemistry-sourabh , I see that you merged #41 but can you raise a PR with unit tests? With out them, they cannot go to master branch(or whatever the plan is for master branch)?

chemistry-sourabh:
We need to sort out the testing framework first. I am interested in something like one file which grabs all test cases  and runs them. I wrote code for something like that, but no response

ravisantoshgudimetla:
I don't know if this is for some discussion but usually - it should be something like a single python file for every unit test available and wrap all up in a shell script or other python script. Thats how we can get started on working atleast. I am not sure, if you have some framework in mind for that. JUnit could be used but its an unnecessary mess at this point. Lets try to have something very simple. 

\cc @sirushtim @naved001 @apoorvemohan

chemistry-sourabh:
I would ask you guys to review the framework what I wrote first before going forward. If that is fine then we can get the existing tests working with travis.

sirushtim:
Just an update for those following this thread, I've updated https://github.com/CCI-MOC/ims/pull/85 with an approach for the unit tests for config files and https://github.com/CCI-MOC/ims/pull/83 for approach on the test framework being used itself. If there are other ideas, it'd be good to hear them before we merge these changes.

chemistry-sourabh:
Merged PR

PR #84, opened 2017-02-26
Config changes

chemistry-sourabh:
Made some changes to config for the tests #69 . Also added some additional checks to improve reliability

### DIFF
--- a/ims/common/config.py
+++ b/ims/common/config.py
@@ -24,7 +24,7 @@ def load(force=False):
                 constants.CONFIG_LOCATION_ENV_VARIABLE]
         except KeyError:
             path = constants.CONFIG_DEFAULT_LOCATION
-        __config = __BMIConfig(path)
+        __config = BMIConfig(path)
         __config.load_config()
         parse_config(__config)
 
@@ -39,10 +39,10 @@ def get():
     return __config
 
 
-class __BMIConfig:
+class BMIConfig:
     def __init__(self, filename):
         self.configfile = filename
-        self.config = None
+        self.config = ConfigParser.SafeConfigParser()
 
     def load_config(self):
         """
@@ -50,10 +50,8 @@ class __BMIConfig:
 
         :return: None
         """
-        config = ConfigParser.SafeConfigParser()
-        if not config.read(self.configfile):
+        if not self.config.read(self.configfile):
             raise IOError('cannot load ' + self.configfile)
-        self.config = config
 
     def option(self, section, option, type=str, required=True):
         """
@@ -73,7 +71,7 @@ class __BMIConfig:
             if section_obj is None:
                 section_obj = ConfigSection()
                 setattr(self, section, section_obj)
-            if type == bool:
+            if type is bool:
                 v = value.lower()
                 if v in ['true', 'false']:
                     setattr(section_obj, option, v == 'true')
--- a/ims/common/config.py
+++ b/ims/common/config.py
@@ -41,7 +41,7 @@ def get():
 
 class BMIConfig:
     """ Contains the config of bmi as attributes """
-    
+
     def __init__(self, filename):
         self.configfile = filename
         self.config = ConfigParser.SafeConfigParser()
--- a/ims/common/config.py
+++ b/ims/common/config.py
@@ -40,6 +40,8 @@ def get():
 
 
 class BMIConfig:
+    """ Contains the config of bmi as attributes """
+    
     def __init__(self, filename):
         self.configfile = filename
         self.config = ConfigParser.SafeConfigParser()
--- a/ims/common/config.py
+++ b/ims/common/config.py
@@ -9,9 +9,16 @@ from ims.common.bmi_config import parse_config
 __config = None
 
 
-def load():
+def load(force=False):
+    """
+    Loads the config by using BMI_CONFIG environment variable, else loads from
+    default location
+
+    :param force: Forces a reload of config
+    :return: None
+    """
     global __config
-    if __config is None:
+    if __config is None or force:
         try:
             path = os.environ[
                 constants.CONFIG_LOCATION_ENV_VARIABLE]
@@ -23,23 +30,43 @@ def load():
 
 
 def get():
+    """
+    Returns the global config
+
+    :return: The Global Config
+    """
     global __config
     return __config
 
 
 class __BMIConfig:
-    # other instance variables are initialized as to not get AttributeErrors
     def __init__(self, filename):
         self.configfile = filename
         self.config = None
 
     def load_config(self):
+        """
+        Loads the config as per the given path
+
+        :return: None
+        """
         config = ConfigParser.SafeConfigParser()
         if not config.read(self.configfile):
             raise IOError('cannot load ' + self.configfile)
         self.config = config
 
     def option(self, section, option, type=str, required=True):
+        """
+        Parses the given option from config, converts to another type if
+        required or raises an exception if found missing and adds it as
+        attribute to config as cfg.section.option
+
+        :param section: Section under which Option should be present
+        :param option: Option should be parsed
+        :param type: the conversion function for the required type like int,etc
+        :param required: Whether an exception should be raised if missing
+        :return: None
+        """
         try:
             value = self.config.get(section, option)
             section_obj = getattr(self, section, None)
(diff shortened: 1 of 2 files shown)


ims/common/config.py
```diff
@@ -23,39 +30,73 @@ def load():
...
         try:
             value = self.config.get(section, option)
             section_obj = getattr(self, section, None)
             if section_obj is None:
                 section_obj = ConfigSection()
                 setattr(self, section, section_obj)
             if type == bool:
-                setattr(section_obj, option, value.lower() == 'true')
+                v = value.lower()
+                if v in ['true', 'false']:
```
naved001: suggestion: `if value.lower() in ['true', 'false']`

chemistry-sourabh: using it below also

chemistry-sourabh:
Yes, tests passed

ims/common/config.py
```diff
@@ -23,39 +30,73 @@ def load():
...
+    """
+    Returns the global config
+
+    :return: The Global Config
+    """
     global __config
     return __config
 
 
 class __BMIConfig:
```
ravisantoshgudimetla: Doc string for this class.

ims/common/config.py
```diff
@@ -23,39 +30,73 @@ def load():
...
+        :param required: Whether an exception should be raised if missing
+        :return: None
+        """
         try:
             value = self.config.get(section, option)
             section_obj = getattr(self, section, None)
             if section_obj is None:
                 section_obj = ConfigSection()
                 setattr(self, section, section_obj)
             if type == bool:
```
ravisantoshgudimetla: nit: Use type is bool as its much more expressive.

ims/einstein/iscsi/tgt.py
```diff
@@ -3,6 +3,7 @@
 import os
```
ravisantoshgudimetla: @chemistry-sourabh Can we ignore reviewing this file?

chemistry-sourabh: This is based of the working tgt commit so review the last commit only

chemistry-sourabh:
@naved001 added some changes have a look

chemistry-sourabh:
@ravisantoshgudimetla finish reviewing this also

chemistry-sourabh:
@ravisantoshgudimetla you missed this

ravisantoshgudimetla:
@chemistry-sourabh ok, I will review this today.

ims/common/config.py
```diff
@@ -9,53 +9,94 @@
 __config = None
```
ravisantoshgudimetla: Why are we using global variables here?

ims/common/config.py
```diff
@@ -9,53 +9,94 @@
...
 
 
-def load():
+def load(force=False):
+    """
```
ravisantoshgudimetla: NIT: Change the name to load_config etc.

ims/common/config.py
```diff
@@ -9,53 +9,94 @@
...
         except KeyError:
             path = constants.CONFIG_DEFAULT_LOCATION
-        __config = __BMIConfig(path)
+        __config = BMIConfig(path)
         __config.load_config()
         parse_config(__config)
 
 
 def get():
+    """
```
ravisantoshgudimetla: Same as above.

ims/common/config.py
```diff
@@ -73,5 +114,6 @@ def section(self, section_name, required=True):
 
 
 class ConfigSection:
+    """ A Simple Object to be used to get cfg.section.option """
     def __init__(self):
```
ravisantoshgudimetla: This may not be related to current change, what does this ConfigSection do? - It seems its not having any methods.

ims/common/config.py
chemistry-sourabh: I used config.load everywhere, I think that is descriptive enough

ims/common/config.py
chemistry-sourabh: Same here also

ims/common/config.py
chemistry-sourabh: I needed an object where I can set attributes to. like cfg.iscsi.ip, here iscsi is a ConfigSection obj.

ims/common/config.py
chemistry-sourabh: So that only one config object exists in the entire application. We dont want to keep loading a different object in each module.

chemistry-sourabh:
@ravisantoshgudimetla your turn

ims/common/config.py
ravisantoshgudimetla: Right but do you need a global variable for that is the question. There are many ways to do this but again this is nit, if you have a strong preference of using global, I am ok with that.

ims/common/config.py
ravisantoshgudimetla: So, is this something related to obtaining a dummy object where you can set attributes? If yes, this is something that has to be changed as it confuses the new developers coming in. We can address this separately.

chemistry-sourabh: You got the idea. How do you propose we address it better ?

chemistry-sourabh:
The test cases I raised in the config tests PR have cases that test for exceptions. Also this PR passed the test cases that I wrote. @naved001 is welcome to do some basic testing.

Also @naved001 it is great if you can convert your manual testing into test cases to reduce yours and everybody's burden. Especially the ones I missed like #91

ims/common/config.py
chemistry-sourabh: What are the other ways ?

ravisantoshgudimetla: Try wrapping it in a class or a method(which is unmutable).

ims/common/config.py
ravisantoshgudimetla: You can use something called object to generate dummy objects. I created a gist for you. I didn't test it though. https://gist.github.com/ravisantoshgudimetla/8fe92f88f433ef2ceaee7100c5ea6b70. I don't know the further implications of it from security stand point but thats a good starting point is what I think.

chemistry-sourabh: I thought about using object, didn't knew if it was the right way to do it. What if later we decide to add methods to the ConfigSections ? Can't we just add some docs and let it be ?

ims/common/config.py
chemistry-sourabh: I dont see how will this work ?

Method will create a new object which we dont want.

A Class will need an object that needs to be global.

chemistry-sourabh:
@ravisantoshgudimetla if the only thing you are waiting for is @naved001's tests then give a +1 and add a do not merge label.

ims/common/config.py
ravisantoshgudimetla: >Method will create a new object each time it is called.

Not always, we can define immutable ones. Read about the ones in:

http://stackoverflow.com/questions/1318406/why-is-the-borg-pattern-better-than-the-singleton-pattern-in-python#1318444

Let me know, if you need some help, I can send some gist over next few weeks.

ims/common/config.py
ravisantoshgudimetla: I thought you wanted a dummy object for making changes. If you have intention of using this later, please keep it with some documentation. (but again there are other ways to do this).

ravisantoshgudimetla:
>@ravisantoshgudimetla if the only thing you are waiting for is @naved001's tests then give a +1 and add a do not merge label.

Done.

naved001:
Approved. works perfectly fine.

ims/exception/config_exceptions.py
```diff
@@ -25,3 +25,17 @@ def __init__(self, section):
 
     def __str__(self):
         return "Missing " + self.option + " section in config file"
+
+
+class InvalidValueConfigException(ConfigException):
```
ravisantoshgudimetla: Please provide some doc string to this like what will cause it etc.

ravisantoshgudimetla:
Also, create an issue for this.

ims/exception/config_exceptions.py
naved001: @chemistry-sourabh add the docstring on this one, then this should be ready to merge.

chemistry-sourabh: I missed this

chemistry-sourabh:
@naved001 Added doc strings, now this is also ready to merge

chemistry-sourabh:
@ravisantoshgudimetla this is related to unit tests for config. I'll link this PR with it.

PR #85, opened 2017-02-26
Config tests

chemistry-sourabh:
Issue #69


tests/common/config/unit/test_config.py
```diff
@@ -0,0 +1,144 @@
...
+
+    @trace
+    def setUp(self):
+        self.old_path = os.getenv('BMI_CONFIG')
+        os.environ['BMI_CONFIG'] = "tests/common/config/config1.cfg"
+
+    def runTest(self):
+        config.load(force=True)
+        cfg = config.get()
+        self.assertEqual(cfg.tests.var1, '0')
```
naved001: quick question: our config file has to have a `[test]` section with a variable called `var1` just for this test? If that's the case, perhaps we could give it a more descriptive name; but even I can't think of something more appropriate right now.

tests/common/config/config5.cfg
```diff
@@ -0,0 +1,58 @@
...
+
+# this section is for logs
+[logs]
+path = /home/bmi/logs/
+debug = True
+verbose = False
+
+[tests]
+var1 = 0
+var2 = 1
```
naved001: we never use `var2` so why have it?

tests/common/config/unit/test_config.py
```diff
@@ -0,0 +1,144 @@
...
+    @trace
+    def setUp(self):
+        self.old_path = os.getenv('BMI_CONFIG')
+        os.environ['BMI_CONFIG'] = "blahblah.cfg"
+
+    def runTest(self):
+        with self.assertRaises(IOError):
+            config.load(force=True)
+
+    def tearDown(self):
```
xukan: Hi, @chemistry-sourabh , one suggestion here, can we have a base class, which contains interface for setUp, tearDown, so that other specific test classes can implement base class and override method in base class. Also, we don't need repetitively write tearDown function in every test case. What do you think of it?

chemistry-sourabh: How will you pass the argument to setup ?

chemistry-sourabh:
I dont have an issue changing the filenames, but having one file with all errors is not possible as exception will be raised when the parser hits the first error, so we won't know if it is able catch all errors. I am open to any better solutions

chemistry-sourabh:
the operations.py fix is part of the tgt-changes PR, this PR is based of it as the PR fixed some tgt issues. Just review the last commit, the other 2 are raised in other PRs. So once those PRs are merged then this PR can be merged.

Regarding the single config thing. I said in the previous comment that we cant have all wrong options in the same config as the parser will stop after hitting the first wrong one (Kind of a like an interpreter). I am open to a solution if you have one in mind.

sirushtim:
I'd like the config files to be explicit about what exactly is being tested, not BMI related. Since each test is testing explicitly a certain scenario that tests maybe 2-3 lines of the config file, we could just do away entirely with separate config files and just have inline strings that are part of each test case. This would also improve readability of tests ie not going through those large config files.

So to be explicit, mock out the load function to return a string that the test case will provide as an inline string.

chemistry-sourabh:
ok makes sense

ravisantoshgudimetla:
@sirushtim The approach seems reasonable but I think the easier way would be to have a config file and load it(This is a unit test to check loading for both +ve and -ve). Rest of times, instead of having a config file, ConfigParser will actually generate a dictionary, so, we can define custom dictionaries and lists in test suite to replicate that. Let me know, what you guys think.

chemistry-sourabh:
So you want to mock out the configparser read function right ?

chemistry-sourabh:
@sirushtim @ravisantoshgudimetla @naved001  have a look

tests/unit/common/sample.cfg
```diff
@@ -0,0 +1,57 @@
...
+
+# this section is for logs
+[logs]
+path = /home/bmi/logs/
+debug = True
+verbose = False
+
+[tests]
+var1 = 0
+var2 = 1
```
ravisantoshgudimetla: New line

chemistry-sourabh: Done!

ravisantoshgudimetla:
@chemistry-sourabh Which commits you want me to review here?

chemistry-sourabh:
config tests onwards, but update tests is the one with the mock changes so you can review that directly.

tests/unit/common/test_config.py
```diff
@@ -0,0 +1,126 @@
+import unittest
+
+import os
+
+from ims.common import config
+
+config.load()
```
ravisantoshgudimetla: NIT: I prefer having a main method instead of loading externally. Its upto you though.

tests/unit/common/test_config.py
```diff
@@ -0,0 +1,126 @@
...
+        os.environ['BMI_CONFIG'] = "blahblah.cfg"
+
+    def runTest(self):
+        with self.assertRaises(IOError):
+            config.load(force=True)
+
+    def tearDown(self):
+        os.environ['BMI_CONFIG'] = self.old_path
+        config.load(force=True)
+
```
ravisantoshgudimetla: what force=False?

tests/unit/common/test_config.py
chemistry-sourabh: it need to be loaded before trace is imported

tests/unit/common/test_config.py
chemistry-sourabh: force=False is default behavior, the config won't be reloaded again from file if it is false.

tests/unit/common/test_config.py
```diff
@@ -0,0 +1,126 @@
...
+        self.cfg.config.add_section('test')
+        self.cfg.config.set('test', 'var1', 'hello')
+
+    def runTest(self):
+        with self.assertRaises(config_exceptions.InvalidValueConfigException):
+            self.cfg.option('test', 'var1', type=bool)
+
+
+class TestMissingOptionalSection(unittest.TestCase):
+    """ Tests if loading when optional section is missing """
```
ravisantoshgudimetla: NIT: Grammar.

chemistry-sourabh: ok

tests/unit/common/test_config.py
```diff
@@ -0,0 +1,126 @@
...
+    @trace
+    def setUp(self):
+        self.cfg = config.BMIConfig("")
+        self.cfg.config.add_section('test')
+        self.cfg.config.set('test', 'var1', 'hello')
+
+    def runTest(self):
+        self.cfg.option('test', 'var1')
+        self.cfg.option('test', 'var2', required=False)
+        self.assertIsNone(getattr(self.cfg.test, 'var2', None))
```
ravisantoshgudimetla: Overall this looks good. One thing that I am not able to understand is why do you want to have separate class for everything? Why can't have a single dictionary with all those values for different sections and get errors in individual functions.

tests/unit/common/test_config.py
ravisantoshgudimetla: So, shouldn't we test that scenario then? We need to have a test case with content being same or am I missing something?

chemistry-sourabh: The force thing is for the test cases. Not used in actual code. It can't be as we are loading the config at the top for trace to work.

tests/unit/common/test_config.py
chemistry-sourabh: Thats how unittest library works. Each Case is a class with a setup and teardown. If some cases have same setup and teardown they can be bundled in same class, but I dont want to do that as it breaks pattern. The Parser might use a dictionary internally it doesn't expose it outside, so better stick to the regular methods.

tests/unit/common/test_config.py
ravisantoshgudimetla: Ok, if you have force as bool variable in code, you should check both in test cases, if you can move the config in some other function and start calling the function after this has been loaded, will there be a problem?

chemistry-sourabh: You mean the initial config.load() at the top ?

tests/unit/common/test_config.py
ravisantoshgudimetla: > but I dont want to do that as it breaks pattern

What pattern are we talking about here. 

> The Parser might use a dictionary internally it doesn't expose it outside, so better stick to the regular methods.

True, I am not asking you to write a dictionary on your own, if you want to set this up, set all these values at once and then do comparisons in various methods.

tests/unit/common/test_config.py
ravisantoshgudimetla: Yes, loading the config file.

tests/unit/common/test_config.py
chemistry-sourabh: 1 case per class. Then the config must be set outside the class if pattern need to be preserved which is also not right.

tests/unit/common/test_config.py
chemistry-sourabh: To test force=False scenario we can load the sample.cfg, change to the old config in the test and try loading with force=False if it changes then fail. What do you think ?

ravisantoshgudimetla: I am fine with either of approaches but what wanted was that variable to be covered as part of test coverage.

chemistry-sourabh: ok will update tomorrow. Stay Tuned!

tests/unit/common/test_config.py
chemistry-sourabh: Unfortunately unittest adds a lot of boilerplate, this is why pytest framework was designed. Pytest just uses functions instead of classes.

ravisantoshgudimetla: What if you use text-fixture where you can initialize this whole structure and use it functions?

chemistry-sourabh: probably a module fixture might do the trick. Will try today evening.

chemistry-sourabh: Tried adding a fixture module, but pytest doesn't run it. So undoed it to have compatability with pytest.

tests/unit/common/test_config.py
chemistry-sourabh: Done

tests/unit/common/test_config.py
chemistry-sourabh: Done

chemistry-sourabh:
@ravisantoshgudimetla have a look

tests/unit/common/test_config.py
ravisantoshgudimetla: There is no need to use fixture, I told one of the ways to do it. You can do it using a normal function as well.

chemistry-sourabh:
Wait you meant the one config object right ?

chemistry-sourabh:
@sirushtim and @ravisantoshgudimetla Have a look

naved001:
@ravisantoshgudimetla @sirushtim 

can you see if your concerns have been addressed here?

chemistry-sourabh:
@ravisantoshgudimetla you are left

tests/unit/common/test_config.py
```diff
@@ -0,0 +1,105 @@
...
+        with self.assertRaises(IOError):
+            config.load(force=True)
+
+    def tearDown(self):
+        os.environ['BMI_CONFIG'] = self.old_path
+        config.load(force=True)
+
+
+class TestCorrectPath(unittest.TestCase):
+    """ Tests if able to load from correct path """
```
ravisantoshgudimetla: Grammatical mistake I guess, nit again.

tests/unit/common/test_config.py
```diff
@@ -0,0 +1,105 @@
...
+    def test_missing_optional_section(self):
+        """ Tests loading a config which has a missing optional section """
+        self.cfg.section('missing', required=False)
+        self.assertIsNone(getattr(self.cfg, 'test', None))
+
+    def test_missing_optional_option(self):
+        """ Tests loading a config which has a missing optional option """
+        self.cfg.option('test', 'var1')
+        self.cfg.option('test', 'var2', required=False)
+        self.assertIsNone(getattr(self.cfg.test, 'var2', None))
```
ravisantoshgudimetla: I may be missing something here but why don't we have a tearDown here?

chemistry-sourabh: @ravisantoshgudimetla It is in memory right the cfg object wont exist outside the test class. Also we are creating a new config object in setUp so it acts like a tearDown.

ravisantoshgudimetla: I meant in the class.

chemistry-sourabh: Ya my point is no need as there is nothing to tearDown

chemistry-sourabh:
@ravisantoshgudimetla Done. @apoorvemohan this is also ready to merge

PR #93, opened 2017-03-24
Docker for BMI

chemistry-sourabh:
Docker scripts for BMI and HIL for dev environment


chemistry-sourabh:
@CCI-MOC/bmi-breakers try it out!!

ravisantoshgudimetla:
@chemistry-sourabh Create a wiki instead of PR.

chemistry-sourabh:
Did you mean the Readme file as a wiki ? or the Dockerfile ? because making the latter a wiki doesn't make sense

ravisantoshgudimetla:
Either way is fine but PR doesn't make sense.

chemistry-sourabh:
@ravisantoshgudimetla what do you mean doesn't make sense ?

ravisantoshgudimetla:
So, do you want this to be used for setting up BMI, if yes raise an issue and write documentation on how to use it.(This could be in alternate ways to install BMI). Even if this goes into repo, the location has to be decided where this has to stay in repo, without that this PR staying here, doesn't make sense.

chemistry-sourabh:
@ravisantoshgudimetla Ya in the future this can be an easier deployment option for people who use something like kubernetes or docker swarm, etc. For now this just sets up a dev environment for easy deployment on the local system. I am using it on a regular basis and is faster than using a VM on kaizen.

I attached a docker.md with instructions on setting it up. I asked people to give it a try, but nobody did :(, Also typically all the docker files go in the root of the project as docker builds the context of the directories below when building images.

We can later upload it to a MOC docker registry (if it exists)

ravisantoshgudimetla:
You can upload to docker hub as of now, if I am not mistaken, everyone should be able to access it if you give privileges. @akshayakhare has created a openshift registry, if I am not mistaken again, we can try uploading there. There is no need to use docker swarm or kubernetes unless we want the software to be container grade in production usage. We can use it quick development setup. I recommend you to write a clear README and add it to dev guide. I can test this over next weekend. This is really cool.

naved001:
@ravisantoshgudimetla should we have this in an `examples/bmi-docker/` directory in the main repo?

chemistry-sourabh:
@naved001 add what to examples ?

naved001:
> add what to examples ?

The scripts in this PR aren't really part of the main BMI code which is why I think it should be in a different folder.

ravisantoshgudimetla:
Let's add a contrib folder in root and tell developers that this is supported way to do development.

chemistry-sourabh:
I am planning on writing a start dev env script using docker-compose (which I learnt about recently). Also I am thinking of submitting a better version of the run tests script to this PR.

@ravisantoshgudimetla @naved001 let me know more stuff which may be required.

apoorvemohan:
@chemistry-sourabh can we close this?

PR #95, opened 2017-03-26
Filesystem interface initial commit

ravisantoshgudimetla:
Initial commit on filesystem interface. Should address #33.

### DIFF
--- a/ims/interfaces/filesystem.py
+++ b/ims/interfaces/filesystem.py
@@ -1,0 +1,67 @@
+from abc import ABCMeta
+from abc import abstractmethod
+
+
+class FileSystem(object):
+    """
+    An abstract class for the filesystem that needs to be implemented by all
+    filesystems used by BMI. This should have basic operations that all
+    filesystems support. Some of the methods' arguments may be too much
+    modelled on ceph. We can change them once we start supporting other
+    filesystems.
+    """
+    __metaclass__ = ABCMeta
+
+    @abstractmethod
+    def list_images(self):
+        """
+        Lists all the images available in filesystem.
+        """
+
+    @abstractmethod
+    def remove_image(self, image):
+        """
+        Removes the given image from filesystem.
+        """
+
+    @abstractmethod
+    def create_image(self, image, size):
+        """
+        Creates an image in the filesystem with given size.
+        """
+
+    @abstractmethod
+    def snap_image(self, image, snapshot):
+        """
+        Creates a snapshot for the given image.
+        """
+
+    @abstractmethod
+    def remove_snapshot(self, image, snap_name):
+        """
+        Removes the given snapshot for the image.
+        """
+
+    @abstractmethod
+    def write(self, image, data, offset):
+        """
+        Writes the given data to image at given offset.
+        """
+
+    @abstractmethod
+    def clone(self, parent_image, parent_snapshot, image):
+        """
+        Creates a clone of parent_image at parent_snapshot with image as name.
+        """
+
+    @abstractmethod
+    def list_snapshots(self, image):
+        """
+        Lists all the snapshots for image.
+        """
+
+    @abstractmethod
+    def remove_snapshot(self, image, snap_name):
+        """
+        Removes snapshot for given image.
+        """
--- a/ims/einstein/ceph.py
+++ b/ims/einstein/ceph.py
@@ -6,6 +6,7 @@ import os
 import rados
 import rbd
 import sh
+from ims.interfaces.filesystem import FileSystem
 
 import ims.common.constants as constants
 import ims.exception.file_system_exceptions as file_system_exceptions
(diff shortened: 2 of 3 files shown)


chemistry-sourabh:
@ravisantoshgudimetla did you test this ? I think every method in the abstract class should have a pass.

ims/interfaces/filesystem.py
```diff
@@ -0,0 +1,67 @@
...
+        """
+
+    @abstractmethod
+    def list_snapshots(self, image):
+        """
+        Lists all the snapshots for image.
+        """
+
+    @abstractmethod
+    def remove_snapshot(self, image, snap_name):
```
chemistry-sourabh: Why are there two remove_snapshots ?

ravisantoshgudimetla: Yeah, this has to be removed.

ravisantoshgudimetla:
@chemistry-sourabh Not all filesystems need explicit locking for creating snapshots. Some of them do implicit locking like gluster. I guess xfs and butterfs does the samething(implicit locking). As, I mentioned, this class needs to have a bare minimum things needed for implementation.

ravisantoshgudimetla:
@chemistry-sourabh  - Pass is not needed for abstract methods.

chemistry-sourabh:
Ok, but we need the method in the interface because we are calling it in operations. We can make the implementation optional

ravisantoshgudimetla:
> Gene and I want to review the refactoring blueprints before we do any further development.


@apoorvemohan I think this was not discussed last time when we met, so I was proceeding with writing stuff. I think, its easier for all us, if you people can define a template for blueprint. This may be a one time activity but saves lots of time for others. Probably, every release can have some blueprints/proposals and then we can define some milestones based on them.

ravisantoshgudimetla:
>Ok, but we need the method in the interface because we are calling it in operations. We can make the implementation optional.

@chemistry-sourabh So, the contract does not mean we cannot implement other methods. We can have any number of methods in that class which implements as long all the abstract methods are implemented. Why to have something which is too much specific in high level abstract class

This is where multi-level inheritence comes in I guess, if you want, you can implement another class which implements basefilesystem and then ceph specific ones(although I recommend against it).

chemistry-sourabh:
My point is if we implement a GlusterFS Class from your interface then it will not work as snap_protect function won't be present in GlusterFS object. It will throw an exception.

Whatever methods the code that uses the object calls should be there in the interface or we should combine some methods.

ravisantoshgudimetla:
Again, why should it throw an error? The interface need to have only methods that are acceptable to all filesystems.

chemistry-sourabh:
Ohho you are not getting me. GlusterFS implements your Interface due to which it doesn't implement snap_protect. Then you call create_snapshot which has a call to snap_protect. Here the fs object will be GlusterFS's object, since it doesn't have snap_protect method it will result in a AttributeError Method not found.

ravisantoshgudimetla:
I think you are thinking too much on the implementation part. Its just a piece of code, you can introspect at run time which tells what object it is and choose a different code path.(It may add a runtime overhead but its always better than making class to implement something which doesnot exist).

chemistry-sourabh:
Atleast you got my point. We need to change the code or interface if the code should work with it.

ravisantoshgudimetla:
>Atleast you got my point. We need to change the code or interface if the code should work with it.

Yeah, I get that. Thanks for the info though. TBH, I did not think about that. But this is what happens when we write code without interfaces. Trying to modify interfaces to work with existing code instead of doing the other way around. A classical example of how we should not do this :).

chemistry-sourabh:
Finally something on which me and @sirushtim agree on LOL.

ravisantoshgudimetla:
>Also, FWIW, IMO, interfaces can have methods that not all backends implement. 

True, its not a question of "can have" but should it?(Other languages won't, unless implementations are done by abstract classes). The problem with python is - it allows and unless, we document it clearly(along with @abstract-method), it becomes difficult to tell once the codebase grows organically.(I guess, this is what we did for iSCSI.). So, I am kind of against this "Notimplemented" thing but if the general consensus is to have something like that, I am fine.

sirushtim:
@ravig Yeah, it should be done on a case-by-case basis. Just that sometimes it might be necessary to add a specific feature one type of hardware supports and others don't.

chemistry-sourabh:
There are 2 options as I said before:-
* Add the missing required methods as optional in the interface. OR
* Change ceph.py and operations.py so that those methods are not needed

chemistry-sourabh:
You guys choose. I prefer 1 as it is simpler, gives flexibility and reduces chances of adding more bugs when compared to 2.

chemistry-sourabh:
@naved001 I think this can be closed for now until we get clarity on the redesigning

naved001:
@chemistry-sourabh I think this is very generic. Don't know if it's worth closing it. I don't know how long the redesign will take, so I actually wanted to finish this PR and then write a mock driver.

naved001:
closing this because the structure will change with the proposed redesign. please reopen if this still has useful code.

Issue #91, opened 2017-03-20
Sanity Check Before Performing CEPH Operations


naved001:
We should first check if the image name exists in ceph or not and then thrown an appropriate error. 
 - Similar to #48

chemistry-sourabh:
Can you describe in what particular case you are getting this bug ?

apoorvemohan:
e.g. When we do node provisioning, in the arguments we specify the golden image name, that is to be used for provisioning the node. Before actually performing the provision operation, we should verify if the image actually exists in BMI database and CEPH. Currently, BMI throws an exception if the image is not found and the database changes are not rolled back leaving the BMI database in inconsistent state.

chemistry-sourabh:
Looked through the code. Yes this case is not handled and needs to be. Right now the db is just returning None if the entry is not found.

naved001:
fixed it by raising an appropriate error when image is not found. i want to make a PR with an accompanying test that reproduces this issue. So I am waiting on #81 to get merged before I do that.

naved001:
We were already performing a check if the image exists or not, but now in #103 we raise an appropriate error.

Issue #57, opened 2017-01-26
Failed BMI provisions can lead to an inconsistent state in BMI


sirushtim:
ubuntu@test-ceph:~/ims/ims$ bmi pro bmi_infra some_bmi_node bmi_image bmi_network 1
bmi_image not found

ubuntu@test-ceph:~/ims/ims$ bmi dpro bmi_infra some_bmi_node bmi_network 1
Got status code 400 from HaaS with message : The network is not attached to the nic.

The issue is likely to be at the tear-down/roll-back process for failed BMI provisions.

chemistry-sourabh:
In most of the cases BMI is not left in an inconsistent state due to the rollback process. But it will happen if the rollback process is screwed up!!

chemistry-sourabh:
It is important to identify what all are the cases where the rollback process fails and possible contingency plans. Also I always wanted to make a reset operation which sort of brings BMI to a consistent state. If we shift to postgres we can probably rollback to the previous commit ?

sirushtim:
> In most of the cases BMI is not left in an inconsistent state due to the rollback process. But it will happen if the rollback process is screwed up!!

Right, BMI does go to an inconsistent state because the roll back process is inherently flawed.

> It is important to identify what all are the cases where the rollback process fails and possible contingency plans. Also I always wanted to make a reset operation which sort of brings BMI to a consistent state. If we shift to postgres we can probably rollback to the previous commit ?

Sure, besides changing a row in a database, we also need to check for cases where we handle stale resources that BMI creates and is not handled appropriately.

Also, for what it's worth, trying to roll back to a previous commit is going to create a long running transaction in the database by sqlalchemy while we wait for the creation of other resources making it inefficient. I'd much rather would like to see a state machine model setting the node to an error state should something happen while we create and letting the user know what happened instead of rolling back and deleting the row entirely. IMO, that's a better user experience than just deleting the node if something went wrong.

chemistry-sourabh:
Your idea sounds more like what openstack does which makes sense and will also reduce a lot of code. But in openstack if the delete of something like an image results in failure., the image doesn't go into a failure state. 

Also if the provision fails and it ends up in the error state how will the user undo it ? By deproing it ? in that case our depro should be made more robust. What if the depro fails ? Should the user continue deproing or proing ? To avoid these confusions it is better to do a reset which removes the node from the error state.

I like this idea so this what I propose.
pro just tries to provision and if fails then ends up in an error state showing the error.
[...]

PR #103, opened 2017-04-13
Raises appropriate error when image is not found

naved001:
Addresses #91 and a little bit of #57 

Now if the image name we give is not in the database, we handle it correctly.

### DIFF
--- a/ims/database/image.py
+++ b/ims/database/image.py
@@ -141,8 +141,11 @@ class ImageRepository:
             image = self.connection.session.query(Image). \
                 filter(Image.project.has(name=project_name)).filter_by(
                 name=name).one_or_none()
-            if image is not None:
-                return image.id
+
+            if image is None:
+                raise db_exceptions.ImageNotFoundException(name)
+
+            return image.id
         except SQLAlchemyError as e:
             raise db_exceptions.ORMException(e.message)
 
--- a/ims/database/image.py
+++ b/ims/database/image.py
@@ -137,7 +137,13 @@ class ImageRepository:
     # returns a array of image ids of the images which have the given name
     @log
     def fetch_id_with_name_from_project(self, name, project_name):
-        """searches for image by name and returns image id"""
+        """
+        searches for image by name and returns image id
+
+        :param  name: name of the image
+        :poaram project_name: name of the project
+        :return: image.id: the image id of the image
+        """
 
         try:
             image = self.connection.session.query(Image). \
--- a/ims/database/image.py
+++ b/ims/database/image.py
@@ -138,11 +138,11 @@ class ImageRepository:
     @log
     def fetch_id_with_name_from_project(self, name, project_name):
         """
-        searches for image by name and returns image id
+        Searches for image by name and returns image id
 
-        :param  name: name of the image
-        :poaram project_name: name of the project
-        :return: image.id: the image id of the image
+        :param name: name of the image
+        :param project_name: name of the project
+        :return: the id of the image.
         """
 
         try:
--- a/ims/database/image.py
+++ b/ims/database/image.py
@@ -137,14 +137,14 @@ class ImageRepository:
     # returns a array of image ids of the images which have the given name
     @log
     def fetch_id_with_name_from_project(self, name, project_name):
+        """searches for image by name and returns image id"""
+
         try:
             image = self.connection.session.query(Image). \
                 filter(Image.project.has(name=project_name)).filter_by(
                 name=name).one_or_none()
-
             if image is None:
                 raise db_exceptions.ImageNotFoundException(name)
-
             return image.id
         except SQLAlchemyError as e:
             raise db_exceptions.ORMException(e.message)
--- a/tests/unit/database/test_image.py
+++ b/tests/unit/database/test_image.py
@@ -6,7 +6,6 @@ config.load()
 from ims.common.log import trace
 from ims.database.database import Database
 from ims.exception import db_exceptions
-import pytest
 
 
 class TestInsert(TestCase):
@@ -71,7 +70,7 @@ class TestFetch(TestCase):
         self.db.image.insert('image 3', 1, parent_id=1)
         self.db.image.insert('image 4', 1, is_snapshot=True, parent_id=1)
 
-    def runTest(self):
+    def test_image_fetch(self):
         images = self.db.image.fetch_images_from_project('project 1')
         self.assertTrue('image 1' in images and 'image 2' in images)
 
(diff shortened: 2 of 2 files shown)


chemistry-sourabh:
@naved001 I thought you were going to write a test for this ? Also are there any other similar cases in db ?

naved001:
@chemistry-sourabh Yes. I'll add it in the next commit asap.

naved001:
> Also are there any other similar cases in db ?

Not sure where else we are missing those, deed to go through it more carefully. But this particular one was causing trouble so I wanted to get a fix out for it soon. And other general cases are being discussed here #57

ims/database/image.py
```diff
@@ -141,8 +141,11 @@ def fetch_id_with_name_from_project(self, name, project_name):
             image = self.connection.session.query(Image). \
                 filter(Image.project.has(name=project_name)).filter_by(
                 name=name).one_or_none()
-            if image is not None:
-                return image.id
+
```
ravisantoshgudimetla: Some additional blank lines. Please remove them. Also, since this is clean up, adding doc strings, unit tests, copy right terms when we are touching files helps a lot.

naved001: This is not really a clean up. Without the changes that I am making here, the db is left in an inconsistent state. I would like to keep this PR short because if I add more stuff (dosctrings, and restructuring) that would delay it's approval. I'll do the cleanup in another PR.

chemistry-sourabh: @naved001 You can just add doc_string to the function you tweaked. If there was a doc_string you would have needed to update it anyways. Also follow the pattern in Config Changes PR for the doc_string

naved001:
I have added a docstring for the function I am modifying here, and then added a small test for it.

tests/unit/database/test_image.py
```diff
@@ -89,6 +91,10 @@ def runTest(self):
                                                                'project 1')
         self.assertEqual(img_id, 2)
 
+        with pytest.raises(db_exceptions.ImageNotFoundException):
```
chemistry-sourabh: Dont use pytest for writing tests, it is being used as a TestRunner only. Use assertRaises instead

tests/unit/database/test_image.py
```diff
@@ -5,6 +5,8 @@
 
 from ims.common.log import trace
 from ims.database.database import Database
+from ims.exception import db_exceptions
+import pytest
```
chemistry-sourabh: Remove

ims/database/image.py
```diff
@@ -137,12 +137,15 @@ def move_image(self, src_project_name, name, dest_project_id,
     # returns a array of image ids of the images which have the given name
     @log
     def fetch_id_with_name_from_project(self, name, project_name):
+        """searches for image by name and returns image id"""
```
chemistry-sourabh: Describe the Args also. Follow Config Changes PR.

tests/unit/database/test_image.py
```diff
@@ -89,6 +91,10 @@ def runTest(self):
                                                                'project 1')
         self.assertEqual(img_id, 2)
 
+        with pytest.raises(db_exceptions.ImageNotFoundException):
+            img_id = self.db.image.fetch_id_with_name_from_project(
+                'some_', 'project 1')
```
chemistry-sourabh: Also when I initially raised my tests PR the tests here were not split to keep the number of classes small. Since we now agreed to use functions in the test case class it would be better to write each test/assert as a separate function.

It will be better if you can split all the tests in this particular class or atleast your test.

ims/database/image.py
```diff
@@ -137,12 +137,21 @@ def move_image(self, src_project_name, name, dest_project_id,
     # returns a array of image ids of the images which have the given name
     @log
     def fetch_id_with_name_from_project(self, name, project_name):
+        """
+        searches for image by name and returns image id
```
sirushtim: NIt: Capitalize the first letter here for a multi-line docstring.

ims/database/image.py
```diff
@@ -137,12 +137,21 @@ def move_image(self, src_project_name, name, dest_project_id,
...
+
+        :param  name: name of the image
+        :poaram project_name: name of the project
```
chemistry-sourabh: Also :poaram

tests/unit/database/test_image.py
```diff
@@ -94,6 +95,14 @@ def runTest(self):
         self.assertEqual([image[1] for image in images],
                          ['image 1', 'image 2', 'image 3', 'image 4'])
 
+    def test_nonexistent_image_fetch(self):
+        """tries to retrieve the image id of an image that doesn't
```
sirushtim: Ditto. Also, the sentences for a Multi-line docstrings are supposed to start on a new line and always end with a period. For example:

"""
Your multi-line
doc-string.
""""

More on that here: https://www.python.org/dev/peps/pep-0257/

ims/database/image.py
```diff
@@ -137,12 +137,21 @@ def move_image(self, src_project_name, name, dest_project_id,
...
+
         try:
             image = self.connection.session.query(Image). \
                 filter(Image.project.has(name=project_name)).filter_by(
                 name=name).one_or_none()
-            if image is not None:
-                return image.id
+            if image is None:
+                raise db_exceptions.ImageNotFoundException(name)
+            return image.id
```
sirushtim: Nit: A return statement placed here could potentially be a bug someday because Python always executes the finally block in a try-except-else-finally construct. So it's good practice to just place the return statement either outside to make it explicit or in the finally section.

naved001: there's no `finally` block here though. why would that be an issue then?
I am returning in the try section, but if try fails we raise an exception.

sirushtim: I mean it could only potentially be a bug, it's actually fine now. It's just good practice is all for the bug I mentioned above. It was just a suggestion though, so feel free to ignore it.

chemistry-sourabh: We can deal with it when we have a finally block, We will need to change in several places otherwise

ims/database/image.py
```diff
@@ -137,12 +137,21 @@ def move_image(self, src_project_name, name, dest_project_id,
     # returns a array of image ids of the images which have the given name
     @log
     def fetch_id_with_name_from_project(self, name, project_name):
+        """
+        Searches for image by name and returns image id
+
+        :param  name: name of the image
+        :param project_name: name of the project
+        :return: image.id: the image id of the image.
```
chemistry-sourabh: No need for image.id to be here

tests/unit/database/test_image.py
```diff
@@ -94,6 +95,15 @@ def runTest(self):
         self.assertEqual([image[1] for image in images],
                          ['image 1', 'image 2', 'image 3', 'image 4'])
 
+    def test_nonexistent_image_fetch(self):
+        """
+        Tries to retrieve the image id of an image that doesn't
+        exist in the db. Should raise an error"""
```
chemistry-sourabh: The last triple quotes should be in new line

naved001: addressed both comments.

ims/database/image.py
```diff
@@ -137,12 +137,21 @@ def move_image(self, src_project_name, name, dest_project_id,
     # returns a array of image ids of the images which have the given name
     @log
     def fetch_id_with_name_from_project(self, name, project_name):
+        """
+        Searches for image by name and returns image id
+
+        :param  name: name of the image
```
chemistry-sourabh: small nit: extra space between param and name. Otherwise good

Issue #79, opened 2017-02-24
All logs do not get appended


chemistry-sourabh:
Sometimes for some unknown reason logs do not get appended when something goes wrong.

ravisantoshgudimetla:
Example and how to reproduce it would be helpful.

chemistry-sourabh:
Since @naved001 was able to reproduce this, I will describe what I know.

If some exception is raised which is not caught by BMI like RegistrationFailed or IOError, OSError,etc the logs are printed. This is because the logger typically buffers the logs and writes them after a while.

To fix it we must either catch all exceptions in rpc_server (I Don't Like this as we have to catch Exception) or disable buffering for logger, But the exception that caused the crash won't be printed, except in the console. When we make BMI as service the latter will be an issue as typically stdout is redirected to /dev/null for services.

chemistry-sourabh:
@naved001 please describe the exact case.

naved001:
@chemistry-sourabh I reproduced this when I deleted the log file. So a new log file was created but logs weren't being written at all.

chemistry-sourabh:
@naved001 Did something crash at that time ?

pgrosu:
@naved001 Did you get a chance to look into this?  If it's still an issue, could you show us how to replicate it to help with tracing it out.

Thanks,
~p

naved001:
@pgrosu last time I checked I couldn't reproduce this issue. Ill have a look at it again.

pgrosu:
@naved001 That sounds good, and agree with the double-check.  Thank you for looking into it.

naved001:
So far I have not been able to reproduce this issue. 

> If some exception is raised which is not caught by BMI like RegistrationFailed or IOError, OSError,etc the logs are printed

We are now catching these issues (https://github.com/CCI-MOC/ims/pull/116), so maybe that's why we don't run into logging issues. 

@apoorvemohan @ravisantoshgudimetla @chemistry-sourabh @sirushtim should we close this issue then? and what about #106 where we are catching the broad exceptions?

pgrosu:
Not yet, still looking into it.

chemistry-sourabh:
@pgrosu were you able to reproduce this ?

pgrosu:
@chemistry-sourabh Sure let's close it.  How are guys changing the verbosity?

naved001:
[...]

PR #106, opened 2017-04-18
Logs Appended Issue

chemistry-sourabh:
This is to address Issue #79 

I didn't test this, but as per the theory I had I am 90% sure this should fix it if I was right. @naved001 I leave it to you to test the issue properly.

Also I tested it against the existing test cases and it passed

### DIFF
--- a/ims/rpc/server/rpc_server.py
+++ b/ims/rpc/server/rpc_server.py
@@ -15,6 +15,14 @@ class MainServer:
     # The commandline arguments following that are the arguments to the method.
     @log
     def execute_command(self, credentials, command, args):
+        """
+        Executes the given BMI command
+
+        :param credentials: The credentials that BMI will use to authenticate.
+        :param command: The BMI command to execute
+        :param args: The Arguments which should be given to BMI Command
+        :return: a dict as { HTTP status code, Output or Error Message }
+        """
         try:
             with BMI(credentials) as bmi:
                 method_to_call = getattr(BMI, command)
@@ -25,6 +33,10 @@ class MainServer:
         except BMIException as ex:
             logger.exception('')
             return {constants.STATUS_CODE_KEY: ex.status_code,
+                    constants.MESSAGE_KEY: str(ex)}
+        except Exception as ex:
+            logger.exception('')
+            return {constants.STATUS_CODE_KEY: 500,
                     constants.MESSAGE_KEY: str(ex)}
 
     @log


ims/rpc/server/rpc_server.py
```diff
@@ -26,6 +34,10 @@ def execute_command(self, credentials, command, args):
             logger.exception('')
             return {constants.STATUS_CODE_KEY: ex.status_code,
                     constants.MESSAGE_KEY: str(ex)}
+        except Exception as ex:
```
ravisantoshgudimetla: Why broader exception here? Any specific reason?

chemistry-sourabh:
@ravisantoshgudimetla my hunch on why it happens is when an exception which is not caught occurs, it crashes logger which results in no logs being written. I placed the broad exception here as it is the entry point. Also another advantage is that the user will get some feedback if something out of the blue happens

ravisantoshgudimetla:
> my hunch on why it happens is when an exception 

Can you test this, if not, its not worth the change.

>Also another advantage is that the user will get some feedback if something out of the blue happens

U mean end-user or developer? If developer, it makes sense, what will happen to end-user, is he/she going to see weird error like "FileNotFound" kind of exception?

ims/rpc/server/rpc_server.py
```diff
@@ -26,6 +34,10 @@ def execute_command(self, credentials, command, args):
...
+            logger.exception('')
+            return {constants.STATUS_CODE_KEY: 500,
+                    constants.MESSAGE_KEY: str(ex)}
```
ravisantoshgudimetla: Should we return str(ex) directly here? Some failures may not make much sense to end-user.

chemistry-sourabh:
The end user will see the message of the exception. I feel it is better than seeing a stack trace. Also @naved001 tried reproducing the issue, but failed. Will try myself later.

ravisantoshgudimetla:
> The end user will see the message of the exception.

Thats weird. Imagine end-user seeing IOException or Security exception. I believe UX will be terrible because of that and honestly, if I am end-user I would be frustrated with exceptions thrown at me that I won't even understand.

ravisantoshgudimetla:
> I feel it is better than seeing a stack trace

Are we returning stack trace now?

chemistry-sourabh:
No. It just crashes with a pointless stack trace. Pyro4 doesnt let us see the actual exception so the message is better.

ravisantoshgudimetla:
>No. It just crashes with a pointless stack trace. Pyro4 doesnt let us see the actual exception so the message is better.

So, the issue is we are not able to overwrite Pyro4's msg and we are intending to send an exception directly? TBH, its like we are solving one problem and creating a new one but not addressing underlying problem.

chemistry-sourabh:
The client side just crashes by showing some pyro4 exception which is pointless for us. We are just sending the message not the exception. Also what new problem are we creating ?

ravisantoshgudimetla:
Irrespective of we catch exception here, it seems end-user receives Pyro4 Communication Exception and 500 as status code(which is ok but we should change it as well and beyond scope of what you are doing in this PR).

>We are just sending the message not the exception.

You are converting the exception to string and sending it back, isn't it?

chemistry-sourabh:
Catching the exception should fix the issue. The return thing is a side effect.

pgrosu:
@naved001 Same comment as https://github.com/CCI-MOC/ims/issues/79#issuecomment-304039997

chemistry-sourabh:
@naved001 please post your findings here.

naved001:
In the comment [here](https://github.com/CCI-MOC/ims/issues/79#issuecomment-292850017)

> If some exception is raised which is not caught by BMI like RegistrationFailed or IOError, OSError,etc the logs are printed. This is because the logger typically buffers the logs and writes them after a while. To fix it we must either catch all exceptions in rpc_server (I Don't Like this as we have to catch Exception) or disable buffering for logger, But the exception that caused the crash won't be printed, except in the console. 

I raised these errors that we aren't handling in some of the calls (`raise IOError`) and the logger never crashed and logs were still being written to. The unexpected Error was not written to the logs, obviously, but everything up to that point was. And any subsequent calls were also logged.

chemistry-sourabh:
So the main point is the logs are printing except for the uncaught exception which may be required for debugging. So I'll let you guys decide @CCI-MOC/bmi-breakers

naved001:
@apoorvemohan while we are at it, what do you think of this PR? We have not been able to reproduce that issue anymore and I think we should move on.

apoorvemohan:
I would suggest to close this and include a note in the release notes.
 @pgrosu What's your opinion?

chemistry-sourabh:
@apoorvemohan so you are saying you dont want the catch all block ?

apoorvemohan:
@chemistry-sourabh No. I thought we have already added the catch and but we are unable to reproduce the error?

chemistry-sourabh:
@naved001 seems patched a catch hole in one of his PRs and was not able to reproduce the error, but if some non caught exception occurs then everything except the error is being printed which I think is important for debugging. It doesnt happen with this PR as I am catching everything.

pgrosu:
So I created the following test, which I ran 1000 times and got no errors:

https://gist.github.com/pgrosu/0ddabda6d8f494bc22999503284accb0

Let's merge this PR as-is, and then use another proxy service approach that is not RPC-based in the redesign that will capture possible failures.

Thanks,
Paul

naved001:
@chemistry-sourabh you want to rebase this so that we have updated travis here?

pgrosu:
+1

chemistry-sourabh:
@naved001 rebased.
@pgrosu could you please use the approve button whenever you want to do a +1 as it will be visible from other web pages

naved001:
how often would we have to deal with the decreasing coverage?

chemistry-sourabh:
@naved001 We can write a test that will cover the 3 new lines added (As per report). The thing is I have no clue on how to do this, as the exception should be generated from within operations.py and should not be caught.

We can probably ignore this and merge it unless someone has a way to write a test for it ?

naved001:
Note: although this does not solve the original issue this was raised for, it allows us to log any unhandled errors/exceptions  which is why we merged it.

Issue #107, opened 2017-04-18
Tests should use in memory db


chemistry-sourabh:
All the tests use the sqlite db specified in the config, which is bad as it overlaps with the prod db. I tried doing it earlier, but had issues.

I finally solved it and it is related to #35

Issue #35, opened 2017-01-10
DB Engine


chemistry-sourabh:
Currently the sqlalchemy's engine is a static variable in the class, but it is being created each time a request is being made which is not good.  
My idea is to write a wrapper that wraps the sqlalchemy's engine.
This class will be directly used in RPC Server and the DBConnection object given by this class will be passed to BMI object.
This will allow us to solve the above issue and allow to change sqlalchemy with another ORM should the need arise.
@ravisantoshgudimetla @apoorvemohan @sirushtim @naved001 give thoughts below

apoorvemohan:
@chemistry-sourabh can we close this?

PR #109, opened 2017-04-19
In memory DB for tests

chemistry-sourabh:
Issues #107 and #35 
Allows to create in memory for tests. Just set db path in config to :memory: and run tests normally.

Also it creates engine only once now.

The current issue is all tests run well except test_rest.py these run against picasso and einstein which create a separate in memory db as they are separate processes and the test process creates a separate db. Any ideas on how to resolve this ?

### DIFF
--- a/ims/database/db_connection.py
+++ b/ims/database/db_connection.py
@@ -4,9 +4,13 @@ from sqlalchemy.orm import sessionmaker
 from sqlalchemy.pool import SingletonThreadPool
 
 import ims.common.config as config
-import ims.common.constants as constants
 
 _cfg = config.get()
+
+# creates the engine for the database
+# Should be adapted to postgres SQL
+engine = create_engine('sqlite:///' + _cfg.db.path,
+                       poolclass=SingletonThreadPool)
 
 
 # The class which represents the BMI database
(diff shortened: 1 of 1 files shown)


chemistry-sourabh:
Ok Sad News!! python guys added the shared cache functionality in python 3. So looks like fixing test_rest.py is not possible in python2. Will still dig a bit, but it doesn't look optimistic.

chemistry-sourabh:
I tried running BMI in python3. Looks like there are not a lot of changes to get BMI running on python3, BUT python-cephlibs doesn't exist for python3 so BMI wont run on python3 unless entire ceph driver is rewritten. So we wont be able to use test_rest.py with In Memory DB.

@CCI-MOC/bmi-breakers Any thoughts ?

naved001:
> So we wont be able to use test_rest.py with In Memory DB.

We can run rest of the tests with an in memory db then I guess. and run `test_rest` with a db on disk. 

Also, just spitballing here, can we modify `test_rest` to create a temporary db (`/tmp`) when it's run and delete it when it finishes or fails.

chemistry-sourabh:
Not a bad idea, but will need to add a new mode in our db for this to work

chemistry-sourabh:
That can be done when the test cases are changed for starting and stopping picasso and einstein

pgrosu:
Not sure if this is the same as @naved001 suggestion, but why not create the database on `/dev/shm` if an in-memory db is required.

chemistry-sourabh:
It wouldnt make a difference as we will still need to create a new db for every test case. If we get sqlite's in memory db to work then it will do the cleanup for us.

pgrosu:
So if every test has its own database instance, and the goal of the in-memory database is for self-cleanup why not create a temporary database via `sqlite3.connect('')`, which will also do its own cleanup.  What are you trying to share via the cache?

chemistry-sourabh:
The cache option is available only in python3 and not in python2 (which is what we use). Since the picasso test cases depend on einstein and picasso instances which have their own db, the only option we have is to use a disk based sqlite db.

ravisantoshgudimetla:
@chemistry-sourabh What is test_rest.py? Is this an integration test or unit test?

chemistry-sourabh:
@ravisantoshgudimetla test_rest.py are the tests that test picasso to einstein part. They can be unit tests if we mock out ceph, hil, etc or integration test as they test the entire end to end thing.

chemistry-sourabh:
@naved001 I am working on a more flexible version of the shell script that I posted on slack. I will raise a PR for it when it is ready. This can also be closed.

PR #116, opened 2017-05-03
Fixes consistency issue

naved001:
* This handles a case where we can get registration failed error due to
permission issues during provision. Allows to roll-back the database to the
correct state.
* may not work if the roll back process itself crashses; @chemistry-sourabh has
a more, hopefully, robust PR that promises to take care of that.

### DIFF
--- a/ims/einstein/operations.py
+++ b/ims/einstein/operations.py
@@ -214,8 +214,8 @@ class BMI:
     @log
     def provision(self, node_name, img_name, network, nic):
         try:
-            mac_addr = ("01-" + self.hil.get_node_mac_addr(node_name).
-                        replace(":", "-"))
+            mac_addr = "01-" + self.hil.get_node_mac_addr(node_name). \
+                replace(":", "-")
             self.hil.attach_node_to_project_network(node_name, network, nic)
 
             parent_id = self.db.image.fetch_id_with_name_from_project(img_name,
--- a/ims/einstein/operations.py
+++ b/ims/einstein/operations.py
@@ -122,9 +122,7 @@ class BMI:
         self.is_admin = self.__check_admin()
 
     @log
-    def __register(self, node_name, img_name, target_name):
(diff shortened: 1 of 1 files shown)


ims/einstein/operations.py
```diff
@@ -216,6 +214,8 @@ def shutdown(self):
     @log
     def provision(self, node_name, img_name, network, nic):
         try:
+            mac_addr = ("01-" + self.hil.get_node_mac_addr(node_name).
+                        replace(":", "-"))
```
chemistry-sourabh: Lol remove extra set of parathesis

naved001: I added that to split it for pep8. Otherwise I don't know how to split that line. want me to use `\`?

chemistry-sourabh: Better

naved001: done

naved001:
@ravisantoshgudimetla There's no test case to be updated. Though I can write one, but I am not sure how I would do that because what I am fixing here only happens when BMI isn't run as root (which leads to IOError or OSError which this handles). 

If you have ideas to write something to test that scenario, then I'll do that otherwise I think it's good.

ravisantoshgudimetla:
Usually, you have to fake it for unit test. But we don't have one yet, easiest would be to update the existing deployment tests that @chemistry-sourabh  wrote some time back, if possible. I just looked at the code, it seems we don't have -ve scenario's being tested. This could be an addition. If not, leave it as it with a TODO, saying that this has to be tested.

naved001:
Given how sporadic my contributions have been, I have decided to do the latter.

chemistry-sourabh:
Better do the later as all the negative test cases, can be written at once

naved001:
@ravisantoshgudimetla made the issue. 

We can just use the squash and merge button :)

chemistry-sourabh:
@naved001 you can do the honors

PR #134, opened 2017-07-11
API for running script on an image

djfinn14:
This API allows SecCloud to run a custom script on an image. BMI is responsible for mapping and unmapping the image.

### DIFF
--- a/ims/cli/cli.py
+++ b/ims/cli/cli.py
@@ -6,6 +6,7 @@ import sys
 import click
 import os
 import requests
+import base64
 from prettytable import PrettyTable
 
 import ims.common.config as config
@@ -87,6 +88,7 @@ def provision(project, node, img, network, nic):
             constants.NIC_PARAMETER: nic}
     res = requests.put(_url + "provision/", data=data,
                        auth=(_username, _password))
+    click.echo(res.url)
     click.echo(res.content)
 
 
@@ -159,7 +161,7 @@ def remove_image(project, img):
 
 @cli.command(name='ls', short_help='List Images Stored')
 @click.argument(constants.PROJECT_PARAMETER)
-def list_images(project):
+def list_image(project):
     """
     Lists Images Under A Project
 
@@ -669,6 +671,47 @@ def download():
     Coming Soon
     """
     click.echo('Not Yet Implemented')
+
+
+@cli.group(short_help='Image Related Commands')
+def image():
+    """
+    Use The Subcommands under this command to manipulate Snapshots
+    """
+    pass
+
+@image.command(name='run_script', help='Run script within BMI')
+@click.argument(constants.PROJECT_PARAMETER)
+@click.argument(constants.IMAGE_NAME_PARAMETER)
+@click.argument(constants.SCRIPT_NAME_PARAMETER)
+@bmi_exception_wrapper
+def run_script(project, img, script):
+    """
+    Run a custom script on a mapped image
+
+    \b
+    Arguments:
+    PROJECT      = The Name of the project
+    IMG          = The Name of the image to run the script on
+    SCRIPT       = The script to run
+    """
+    
+    with open(script, "rb") as f:
+	bin_script = f.read()
+
+    encode_script = base64.b64encode(bin_script)
+
+    data = {constants.PROJECT_PARAMETER: project,
+		constants.IMAGE_NAME_PARAMETER: img,
+		constants.SCRIPT_NAME_PARAMETER: encode_script}
+    res = requests.post(_url + "run_script/", data=data, 
+			auth=(_username, _password))
+
+    json_res = json.loads(res.content)
+    output = base64.b64decode(json_res["stdout"])
+    json_res["stdout"] = output
+
+    click.echo(json_res)
 
 
 if __name__ == '__main__':
--- a/ims/cli/cli.py
+++ b/ims/cli/cli.py
@@ -680,6 +680,7 @@ def image():
     """
     pass
 
+
 @image.command(name='run_script', help='Run script within BMI')
 @click.argument(constants.PROJECT_PARAMETER)
 @click.argument(constants.IMAGE_NAME_PARAMETER)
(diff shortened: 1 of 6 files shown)


ims/common/constants.py
```diff
@@ -124,4 +126,4 @@
 
 BMI_ADMIN_PROJECT = "bmi_infra"
 
-HAAS_BMI_CHANNEL = "vlan/native"
+HAAS_BMI_CHANNEL = "null"
```
naved001: why did you change this to null? that wouldn't work in real life.

ims/cli/cli.py
```diff
@@ -159,7 +161,7 @@ def remove_image(project, img):
 
 @cli.command(name='ls', short_help='List Images Stored')
 @click.argument(constants.PROJECT_PARAMETER)
-def list_images(project):
+def list_image(project):
```
naved001: why change of name here?

ims/einstein/operations.py
```diff
@@ -635,3 +639,31 @@ def remake_mappings(self):
...
+    def run_script(self, img, script):
+        try:
+            if not self.is_admin:
+                raise AuthorizationFailedException()
+            directory = "/tmp/script/"
+            os.makedirs(directory)
+            script_path = directory + "/test_script.sh"
+            ret = self.fs.map(img)
+            decode_script = base64.b64decode(script)
+            with open(script_path, "w") as write_script:
```
chemistry-sourabh: Shouldn't this be wb ? as you did a rb when reading

ims/einstein/operations.py
```diff
@@ -635,3 +639,31 @@ def remake_mappings(self):
...
+                write_script.write(decode_script)
+            os.chmod(script_path, S_IRWXU)
+            p = subprocess.Popen(script_path + " " + ret, shell=True,
```
chemistry-sourabh: This shell=True is dangerous. Also we are just mapping the image, but not mounting it on the VM due to which the script will not be able to access the contents of the image.

The right way to do it would be to mount the image using the "mount" command in some location in /tmp, chroot to that location and then run the script.

Right now if the user is admin he can technically run any script which can be "rm -rf /" which is dangerous. The chrooting will help with this, but might delete the entire image. So it would be better to clone the image before mapping and deleting it at the end.

ims/einstein/operations.py
```diff
@@ -635,3 +639,31 @@ def remake_mappings(self):
...
+                                 stderr=subprocess.STDOUT,
+                                 stdout=subprocess.PIPE)
+            output, err = p.communicate()
+            code = p.returncode
+            encode_output = base64.b64encode(output)
+            return self.__return_success({'stdout': encode_output,
```
chemistry-sourabh: There is no need to send err as it will be included in stdout.

ims/einstein/operations.py
chemistry-sourabh: Also PS mount will not work for all filesystems. You will probably need to do more stuff to support those filesystems.

djfinn14: The design we decided on with the Secure Cloud team has their script performing the actual mounting of the image. I will change it to clone the image before mapping and delete it after.

ims/cli/cli.py
```diff
@@ -671,5 +673,47 @@ def download():
     click.echo('Not Yet Implemented')
 
 
+@cli.group(short_help='Image Related Commands')
+def image():
+    """
+    Use The Subcommands under this command to manipulate Snapshots
```
djfinn14: Since we have moved toward the direction where this is not going to be in core BMI in the long run, but instead would more likely be a separate service, I renamed the group to just be "script" so it will be easier to break it out of the codebase.

chemistry-sourabh:
@djfinn14 could you please comment on the comments that you solved with your new commit so that we can have a look.

ims/cli/cli.py
```diff
@@ -87,6 +88,7 @@ def provision(project, node, img, network, nic):
             constants.NIC_PARAMETER: nic}
     res = requests.put(_url + "provision/", data=data,
                        auth=(_username, _password))
+    click.echo(res.url)
```
djfinn14: Mistake, fixed.

ims/cli/cli.py
djfinn14: fixed

ims/cli/cli.py
djfinn14: changes made

ims/cli/cli.py
```diff
@@ -671,5 +673,47 @@ def download():
...
+    """
+    Use The Subcommands under this command to manipulate Snapshots
+    """
+    pass
+
+
+@image.command(name='run_script', help='Run script within BMI')
+@click.argument(constants.PROJECT_PARAMETER)
+@click.argument(constants.IMAGE_NAME_PARAMETER)
+@click.argument(constants.SCRIPT_NAME_PARAMETER)
```
djfinn14: fixed

ims/common/constants.py
djfinn14: fixed.

ims/einstein/operations.py
```diff
@@ -635,3 +639,31 @@ def remake_mappings(self):
             logger.exception('')
         except NotImplementedError:
             pass
+
+    @log
+    def run_script(self, img, script):
+        try:
+            if not self.is_admin:
+                raise AuthorizationFailedException()
+            directory = "/tmp/script/"
```
djfinn14: made the change.

ims/einstein/operations.py
djfinn14: fixed.

ims/einstein/operations.py
djfinn14: made the change.

naved001:
@djfinn14 ran this for me and it works as expected.

naved001:
The coverage has decreased because we didn't write tests for the new code. But we are making an exception here since it is for secure cloud paper.

I am gonna go ahead and merge this.

Issue #75, opened 2017-02-22
Move __register() and sub functions to DHCP class


chemistry-sourabh:
Currently the __register method generated the .ipxe and mac address files using two other helper functions and they are part of BMI class.

Should they be moved to DHCP class or make a new TFTP class and include them ?

ravisantoshgudimetla:
@chemistry-sourabh please add line numbers and references in github issue. The more information you provide, it becomes much easier to discuss without going back and forth. 

This idea is interesting. TBH, we didn't think about that when writing this piece. I guess first we need to set the expectation from our side straight. Dnsmasq(the one which we currently have , includes both DHCP and TFTP servers) and we need to support that for sure. For most of DHCP servers they don't include TFTP server. So, no point in making a class for that. But I think the functions that generate these files should be part of DHCP server like __generate_ipxe and others. Let me know, what you think.

chemistry-sourabh:
We can add the API calls to the DHCP class as DNSmasq supports it, to support other dhcp servers that dont have TFTP in them, we can create a tftp class for that server and use composition. Let me know if this makes sense to you.

ravisantoshgudimetla:
So, instead of naming it as DHCP+TFTP why can't we have something like class PXEServer. It should always include both DHCP and TFTP? - (We can have an interface for this as well).

chemistry-sourabh:
I guess this makes more sense. From now on Dnsmasq shall be called as PXEServer not DHCP (:P)

chemistry-sourabh:
@ravisantoshgudimetla help wanted ? for what ?

ravisantoshgudimetla:
Means, external contributors can come in and help.

ravisantoshgudimetla:
BTW, do we have unit-tests for them? If not, this becomes a high priority thing for release 0.5
cc @naved001.

ravisantoshgudimetla:
@chemistry-sourabh Do we have unit tests for this? if not, this becomes high priority.

chemistry-sourabh:
@ravisantoshgudimetla This class only has 1 method that gets IP of a node, which is very OS specific. When we move the functions to this class we can add the test cases with it.

ravisantoshgudimetla:
Is this part of milestone 0.5?

chemistry-sourabh:
no @ravisantoshgudimetla

PR #139, opened 2017-07-20
API to get iSCSI target information from a provisioned node

djfinn14:
Created an API for Secure Cloud to get iSCI target information from a provisioned node through the .ipxe file

### DIFF
--- a/ims/cli/cli.py
+++ b/ims/cli/cli.py
@@ -576,31 +576,6 @@ def get_node_ip(project, node):
             click.echo(ret[constants.MESSAGE_KEY])
 
 
-@node.command('iscsi', help='Get iSCSI target info for a provisioned node')
-@click.argument(constants.PROJECT_PARAMETER)
-@click.argument(constants.NODE_NAME_PARAMETER)
-@bmi_exception_wrapper
-def get_iscsi_target(project, node):
-    """
-    Get the iSCSI target information for a provisioned node
-
-    \b
-    Arguments:
-    PROJECT  = The HIL Project attached to your credentials
-    NODE     = The node whose iSCSI target info is required
-    """
-    data = {constants.PROJECT_PARAMETER: project,
-            constants.NODE_NAME_PARAMETER: node}
-    ret = requests.get(_url + "node/iscsi/",
-                       data=data, auth=(_username,
-                                        _password))
-    if ret.status_code == 200:
-        json_ret = json.loads(ret.content)
-        click.echo(json_ret["iscsi_target"])
-    else:
-        click.echo(ret.content)
-
-
 @cli.group(help='ISCSI Related Commands')
 def iscsi():
     pass
@@ -679,6 +654,32 @@ def show_mappings(project):
     #     else:
     #         click.echo(ret[constants.MESSAGE_KEY])
     click.echo("Need to Re-Implement")
+
+
+@iscsi.command('iscsi_target_info',
+               help='Get iSCSI target info for a provisioned node')
+@click.argument(constants.PROJECT_PARAMETER)
+@click.argument(constants.NODE_NAME_PARAMETER)
+@bmi_exception_wrapper
+def iscsi_target_info(project, node):
+    """
+    Get the iSCSI target information for a provisioned node
+
+    \b
+    Arguments:
+    PROJECT  = The HIL Project attached to your credentials
+    NODE     = The node whose iSCSI target info is required
+    """
+    data = {constants.PROJECT_PARAMETER: project,
+            constants.NODE_NAME_PARAMETER: node}
+    ret = requests.get(_url + "iscsi/iscsi_target_info/",
+                       data=data, auth=(_username,
+                                        _password))
+    if ret.status_code == 200:
+        json_ret = json.loads(ret.content)
+        click.echo(json_ret["iscsi_target"])
+    else:
+        click.echo(ret.content)
 
 
 @cli.command(name='upload', help='Upload Image to BMI')
--- a/ims/cli/cli.py
+++ b/ims/cli/cli.py
@@ -592,13 +592,14 @@ def get_iscsi_target(project, node):
     data = {constants.PROJECT_PARAMETER: project,
             constants.NODE_NAME_PARAMETER: node}
     ret = requests.get(_url + "node/iscsi/",
-                            data=data,auth=(_username,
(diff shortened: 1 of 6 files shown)


ims/einstein/operations.py
```diff
@@ -681,3 +682,26 @@ def run_script(self, img, script):
...
+            if not self.is_admin:
+                raise AuthorizationFailedException()
+            provisioned = False
+            provisioned_nodes = self.list_provisioned_nodes()['retval']
+            for node in provisioned_nodes:
+                if node_name == node[0]:
+                    provisioned = True
+            if not provisioned:
+                raise iscsi_exceptions.TargetDoesntExistException()
+            file_path = "/var/lib/tftpboot/" + node_name + ".ipxe"
```
ravisantoshgudimetla: This could be separated into another fn and should be in dnsmasq.py I guess.(this class name should be changed to PXEServer). Refer #75

ims/einstein/operations.py
```diff
@@ -681,3 +682,26 @@ def run_script(self, img, script):
...
+    # TODO: Write tests for this method before merging to main branch
+    @log
+    def get_iscsi_target(self, node_name):
+        try:
+            if not self.is_admin:
+                raise AuthorizationFailedException()
+            provisioned = False
+            provisioned_nodes = self.list_provisioned_nodes()['retval']
+            for node in provisioned_nodes:
+                if node_name == node[0]:
```
ravisantoshgudimetla: Why node[0]?

ims/einstein/operations.py
chemistry-sourabh: Not a bad ide, Also the path is a config setting use that

ims/cli/cli.py
```diff
@@ -576,6 +576,31 @@ def get_node_ip(project, node):
             click.echo(ret[constants.MESSAGE_KEY])
 
 
+@node.command('iscsi', help='Get iSCSI target info for a provisioned node')
+@click.argument(constants.PROJECT_PARAMETER)
+@click.argument(constants.NODE_NAME_PARAMETER)
+@bmi_exception_wrapper
+def get_iscsi_target(project, node):
```
chemistry-sourabh: I think this should be in the iscsi sub command? You can call it info or some other apt name

ims/einstein/operations.py
chemistry-sourabh: Checked the function and it returns a list of [image, parent_image]. You can do something like

```
if node_name in [node[0] for node in provisioned_nodes]:
    pass
```
I guess it is cleaner. Unless @ravisantoshgudimetla has something better in mind.

Also it would be better to change the nested list to a list of named tuples. What do you think @ravisantoshgudimetla ?

ims/einstein/operations.py
```diff
@@ -681,3 +682,26 @@ def run_script(self, img, script):
...
+                    provisioned = True
+            if not provisioned:
+                raise iscsi_exceptions.TargetDoesntExistException()
+            file_path = "/var/lib/tftpboot/" + node_name + ".ipxe"
+            with open(file_path, "r") as ipxe_file:
+                for line in ipxe_file:
+                    if "iscsi" in line:
+                        target = line.split("--keep ")[1][:-1]
+                        return self.__return_success({'iscsi_target': target})
```
chemistry-sourabh: I guess you are just returning the string after --keep here ? Why not parse it and return a dict ? like 

```
{
    'ip': <>,
    'port': <>,
    'lun_id': <>,
    'target_name': <>
}
```

ims/einstein/operations.py
```diff
@@ -681,3 +682,26 @@ def run_script(self, img, script):
                                    constants.DEFAULT_SNAPSHOT_NAME)
             self.fs.remove_snapshot(ceph_img_name,
                                     constants.DEFAULT_SNAPSHOT_NAME)
+
+    # TODO: Write tests for this method before merging to main branch
+    @log
+    def get_iscsi_target(self, node_name):
+        try:
```
chemistry-sourabh: Missing docstring

ims/cli/cli.py
ravisantoshgudimetla: Probably iscsi_target_info

ims/einstein/operations.py
ravisantoshgudimetla: Wait, why are we returning both image and parent name in list_images? I know that we might have done it for a specific reason but I can't remember of anything. Also, this needs to be separate discussion, let's not block PR on this.

ims/einstein/operations.py
djfinn14: The Secure Cloud team wants to parse out the information the need themselves and asked to just return a string for now, atleast.

djfinn14:
@chemistry-sourabh  @ravisantoshgudimetla ready for another round of review.

ravisantoshgudimetla:
At high level, the changes look ok but in general I don't like the idea of having anything specific to iscsi in operations.py. This functions in this file should be on operating on nodes is what I think. That being said, I don't have problem approving for secure cloud branch.

naved001:
@chemistry-sourabh waiting on your review now.

ims/einstein/operations.py
```diff
@@ -681,3 +682,30 @@ def run_script(self, img, script):
...
+
+    # TODO: Write tests for this method before merging to main branch
+    @log
+    def iscsi_target_info(self, node_name):
+        """
+        Given the name of a provisioned node, returns the iSCSI
+        target information for a provisioned node
+
+        :param node_name: Name of the provisioned node
+        :return: Dictionary of iscsi ip, port number, LUN ID, and
```
chemistry-sourabh: Returns iscsi info string containing ip, port number, etc.

ims/einstein/dnsmasq.py
```diff
@@ -10,3 +10,6 @@ def get_ip(self, mac_addr):
                 if parts[1] == mac_addr and parts[4] == '01:' + mac_addr:
                     return parts[2]
             raise dhcp_exceptions.MacAddrNotFoundException(mac_addr)
+
+    def get_ipxe_path(self, cfg, node_name):
```
chemistry-sourabh: Missing doc_string

naved001:
Coverage has decreased, but I'll merge it anyway because it's for secure cloud.

PR #140, opened 2017-07-26
Hil flag for Secure Cloud

djfinn14:
Added the option to not have BMI call HIL to attach/detach the node from the network as Secure Cloud has their own specific way they want this attachment and detachment to occur.

### DIFF
--- a/ims/einstein/operations.py
+++ b/ims/einstein/operations.py
@@ -219,10 +219,12 @@ class BMI:
     @log
     def provision(self, hil_flag, node_name, img_name, network, nic):
         try:
+            hil_bool = bool(hil_flag == "True")
+
             mac_addr = "01-" + self.hil.get_node_mac_addr(node_name). \
                 replace(":", "-")
 
-            if hil_flag:
+            if hil_bool:
                 self.hil.attach_node_to_project_network(node_name, network,
                                                         nic)
 
@@ -250,7 +252,7 @@ class BMI:
             self.fs.remove(clone_ceph_name)
             self.db.image.delete_with_name_from_project(node_name, self.proj)
             time.sleep(constants.HAAS_CALL_TIMEOUT)
-            if hil_flag:
+            if hil_bool:
                 self.hil.detach_node_from_project_network(node_name, network,
                                                           nic)
             return self.__return_error(e)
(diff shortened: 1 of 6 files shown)


chemistry-sourabh:
@CCI-MOC/bmi-breakers did we reach consensus that we will be removing HIL calls ?

If yes, then I think that this PR should do it instead of doing the flag based approach.

ravisantoshgudimetla:
This is the first time, I am looking at it. I don't have much context to  comment on this.

djfinn14:
This was a decision I came to with the Secure Cloud team to have a flag for now so we don't get rid of the current functionality of BMI while still providing them with what they need.

> @CCI-MOC/bmi-breakers did we reach consensus that we will be removing HIL calls ?

If we can't come up with a consensus very soon on this, I propose we go through with reviewing my PR as is because the Secure Cloud team has a deadline of mid-August for their paper and I do not want to be blocking them for that.

naved001:
I am fine with the flag to disable/enable HIL calls if it's only for secure cloud branch. But I don't think it would be a good idea to have this in the main branch later on; we should stick with one way, isn't it?

chemistry-sourabh:
If we agree on removing HIL calls, then this will be redundant work

naved001:
But at this point, we don't know how the redesign is going to be like. So let's not block the secure cloud people on this PR; it's going to a different branch anyway.

naved001:
@chemistry-sourabh do you want to review this?

ims/einstein/operations.py
```diff
@@ -217,11 +217,16 @@ def shutdown(self):
 
     # Provisions from HaaS and Boots the given node with given image
     @log
-    def provision(self, node_name, img_name, network, nic):
+    def provision(self, node_name, img_name, network, nic, hil_flag):
         try:
+            hil_bool = bool(hil_flag == "True")
```
chemistry-sourabh: Why is "True" being passed instead of True ?

djfinn14: It is originally set as a boolean in the CLI, but when it gets passed to operations for some reason it gets converted in the background to a string, so this was the only way I could see to get it functioning appropriately.

chemistry-sourabh: Could you please spend a bit more time in figuring out why it is being passed as string?

If it is due to rpc, then try going through pyro4 lib if there is something which wont convert to string.

chemistry-sourabh: I am guessing in picasso/rest.py. In the _rest_wrapper if we wrap the args with json.loads, then probably that should fix it.

pgrosu: I would not recommend that here, as I would like to have semantic transparency in the redesign with type-safety via message passing where each field is typed.  By the way, there is no json here - it's just unicode for each parameter.  We need to redesign BMI properly from the ground up, and to infer types in the current design is just work that we can put in the redesign.

This is for the cloud project so that they have their paper published in time.

Issue #102, opened 2017-04-12
Development is not happening in the master branch


knikolla:
PRs are being made and merged against the ``dev`` branch instead of ``master``. The ``master`` branch should be where all the development is happening.

chemistry-sourabh:
@knikolla Master branch should always be the stable branch right ? Not the latest

knikolla:
@chemistry-sourabh ``master`` should be the latest. Here's a brief description of how this model works.

Let's say you release a version codenamed ``hero``. This release maps to your 0.2 release. You would create a new branch named ``hero`` from master (which is your latest development branch), and tag the current latest commit in the ``hero`` branch as 0.2. This is your stable branch for this specific release. You can have many stable branches for many releases, depending on your EOL support policy.

Let's say you are working in ``master`` towards the 0.3 release, but it's not quite ready yet. 0.3 contains non-backwards compatible changes, so you can't release it as 0.2.1. However a very important security bug was just discovered and you need to "backport" the fix for it to 0.2.1, because that release is still stable and supported. What happens is that you fix the bug in ``master``, and cherry pick that specific commit into the ``hero`` branch, then tag the current latest commit in ``hero`` as 0.2.1. (Repeat for all supported branches).

Please note, I'm not opposed to development branches for specific features. However that branch must be short-lived and specific to that single feature and must have fully merged back to master and the branch be deleted for it to be released.

chemistry-sourabh:
We were sort of aiming for the gitflow workflow
https://www.atlassian.com/git/tutorials/comparing-workflows#gitflow-workflow

chemistry-sourabh:
Also @knikolla in your approach when will we merge the release branch into master ?

knikolla:
@chemistry-sourabh The release branch is never merged back. Why would you need to merge it back into master?

naved001:
AFAIK, the reason we decided on a `dev` branch was because, the current `master` is dead and the code in production was coming from @chemistry-sourabh's personal github fork. I wanted to the code to be a part of CCI-MOC. But then we couldn't directly merge it to the master here without fixing some really important issues, which is why we made this dev branch. We have some issues tagged as high priority, once we have those solved we will make this `dev` branch the `master` branch and then follow the standard git workdlow. @apoorvemohan you want to add something to it?

apoorvemohan:
The current master branch is dead. The 0.5-release will be made the master branch. It will be released soon.

chemistry-sourabh:
[...]

Issue #122, opened 2017-06-02
Image removed from ceph but not in the database


naved001:
I was importing and removing images from the bmi db and ceph. not sure what went wrong, but the image was removed from ceph and not the database, leaving the db in an inconsistent state.

apoorvemohan:
is this reproducible?

chemistry-sourabh:
@naved001 did you inspect the logs ?

pgrosu:
Shouldn't the `bmi import` command create a clone in the database?  Somehow it thinks it is there, but I do not see it:

```
(ims) [root@rhel-tgt-base ims]# rbd ls | grep bmi-test-image
bmi-test-image
(ims) [root@rhel-tgt-base ims]# bmi import bmi_infra bmi-test-image
snapshot Already Exists
(ims) [root@rhel-tgt-base ims]# bmi db ls
+----+-------------------+-----------+-------+--------+----------+--------+
| Id |        Name       |  Project  |  Ceph | Public | Snapshot | Parent |
+----+-------------------+-----------+-------+--------+----------+--------+
| 1  |     naved-img1    | bmi_infra | 5img1 | False  |  False   |        |
| 3  | centos7-radostyle | bmi_infra | 5img3 | False  |  False   |        |
| 4  |   centos7-naved   | bmi_infra | 5img4 | False  |  False   |        |
| 5  |     paul-image    | bmi_infra | 5img5 | False  |  False   |        |
+----+-------------------+-----------+-------+--------+----------+--------+
(ims) [root@rhel-tgt-base ims]#
```

apoorvemohan:
@naved001 Were you able to do the sanity check that we discussed?

chemistry-sourabh:
@pgrosu sorry for the late reply. The issue is that bmi import first create a snapshot with the name "snapshot" for the image and then does a clone. I am guessing that your image already has such a snapshot. If you delete it, then import should work.

chemistry-sourabh:
@naved001 if there is an officially sanity check then make it into a document so that other people can refer when testing there stuff. Also if this check can be made into test cases then it is even better.

naved001:
@chemistry-sourabh there's no official sanity check AFAIK. Apoorve suggested that sometimes the image may stay in ceph but it's name might get removed. Basically he wanted me to double check in ceph. @apoorvemohan correct me if I'm wrong.

pgrosu:
[...]

PR #143, opened 2017-08-23
Merge dev into master

naved001:
I think we need to change some link in the ReadMe. Besides that, do we need to change anything else?


naved001:
cc: @pgrosu @apoorvemohan

naved001:
Fixes #102

naved001:
had a discussion with @chemistry-sourabh 

we would need to make several changes in the readme. but we can't do this in this PR otherwise that would break the readme in the `dev` branch.

changes are:
  * change link to image
  * link to build and coverage tags on top of readme.

pgrosu:
+1

apoorvemohan:
I would like to see the list of raised issues after running UATs. As UATs have not being reviewed, I would like to verify those issue once.

naved001:
You might wanna look for @pgrosu's comments on previous 4-5 PRs that we merged.

pgrosu:
It is probably most efficient for me to write the manual for the UAT framework, and provide it to the whole team.  This would include the different system configurations, as some of the issues we encountered on different systems were communicated orally, as we were working together to make implementations operational.  Probably the most significant issue this summer - while running and stress-testing with the UAT - was that we learned BMI was using the same Ceph cluster for both production and development (see https://github.com/CCI-MOC/ims/issues/122#issuecomment-308927324), which also made us aware a delay between image creation and bmi-importing is necessary because of how the Ceph architecture reaches consistency.  The scope of the UAT was to ensure that we catch most of the common use cases - through an end-to-end scenario - so that when folks use it in production - or to learn via the BMI User's Guide (chapter 2) - it will not affect our focus on the redesign with such a small team.

apoorvemohan:
Had a discussion with Gene - if we have any issues, we will release a 0.6

PR #150, opened 2017-10-11
Added the BMI User's Guide

pgrosu:
Added the BMI User's Guide


docs/BMI User's Guide/BMI_User's_Guide-blx.bib
```diff
@@ -0,0 +1,11 @@
+@Comment{$ biblatex control file $}
+@Comment{$ biblatex version 1.7 $}
+Do not modify this file!
+
+This is an auxiliary file used by the 'biblatex' package.
+This file may safely be deleted. It will be recreated as
```
naved001: says right here that this could be deleted. why put it on github at all?

docs/BMI User's Guide/address.txt
```diff
@@ -0,0 +1,5 @@
+if [ ! -z "`ifconfig | grep inet | head -n1 | grep :`" ]; then
```
naved001: what's going on here? what IP address are you getting with this?

naved001 (changes requested):
Some inline comments, and I am gonna write another comment because the "review summary" box is too tiny making it difficult to format.

naved001:
I am not happy with the way the documentation is done. It is an unmaintainable
mess; there's no straightforward way to modify or update this in the future. 105 files (with a lot of screenshots) just for having our documentation. I know that this must have taken a lot of time and effort
but I don't think this is feasible/reasonable.

My first gripe with this is why did you choose LaTex. It's an overkill for having documentation in the source tree here. The best use of LaTex could be to make PDFs and then publish only the PDF when we make a release, not the whole LaTex pile.

For documentation on github, you could have used either markdown or rst, and then have it rendered using Sphinx or something to make it pretty (it can then be hosted on, for example, readthedocs). That would have been easier to maintain, read, and edit by newbies since the learning curve is small.

In any case, even if people wanted to have LaTeX here then here are some comments/questions
about that.
1. Why are there so many `.aux` `.tex` `.bib` `.sty` files? Isn't there a simple way to package only the latex source code (or whatever the equivalent of that is)?
2. I think there should be some basic instructions on what to do with those files.  Should I just runt the make file?
2. I couldn't comment on the images so I am pointing it out here.
  * The HIL login page information shouldn't be a part of BMI guidelines
  * It's still using the old name for HIL.
  * one of the screenshots still point to the dev branch
  * why is there a.....
  * screenshot just to show how to activate virtualenv. I think the command should suffice.
  * and there's one showing how to deactivate.
  * screenshot showing how to source the bmi init file is an overkill.
  * blank bmi terminal.
  * blank bmi terminal with vertical split.
  * blank bmi terminal with a vertical split and a green cursor.
  * several screenshots showing the word "success" after running a command. Is a screenshot really needed there?
  * there's critical dark red.png and a bunch of other icon files.
  * there's bunch of more stuff that didn;t get rendered, but I hope you get the point.

I don't know who it is targeted towards. Given how big this is, and if others from the BMI team would want to keep this in it's current form, I would suggest having a different repo just for the documentation. Or at least a separate folder for the LaTeX source files.

I wanted to comment on the actual content of the documents but I dont know how to do that since the pdf is not visible on github, and digging through the latex source code to comment is painful.

chemistry-sourabh (changes requested):
I didnt read the docs, but I agree with @naved001 that the docs should be written in plaintext or Markdown so that they are easy to modify in the future.

Most of us don't know how to write/modify Latex docs and it has a steep learning curve associated with it. 

105 files is too much to review.

I didn't even like the idea of having a pdf as it is difficult to suggest changes in GitHub and people will still need to mess with Latex at the end. Instead you can do something similar to what @naved001 did for HIL with readthedocs. You will get the similar chapter style flow.

Another idea is to transfer all the docs into the repo's wiki. @apoorvemohan @naved001 what do you guys think ? We can have chapters there and everything. It will be easier to traverse compared to regular docs/

PR #154, opened 2017-11-07
Redesign Specs

chemistry-sourabh:
First Draft of BMI Redesign specs


nasibehteimouri:
All is matched with what I learnt from you guys and the discussions. Great job!

knikolla:
There goes my quiet evening!

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+
+* id : Id of the image (primary key)
+* name : Name of the image
+* ownerUserId : Id of the user who owns the image.
+* type : Type of image (e.g. RAW, QCOW2, etc.)
+* visibility : Access level of image (e.g. private, public, shared, etc.)
+* isSnapshot : This flag is `true` in case of snapshot and not an uploaded image
+* dataStoreId : Id of the DataStore that gives information about where this image is stored. 
+
+**NOTE:** `(name, ownerUserId) pair is unique key.`
```
knikolla: If you have an image named `openstack-compute` and I try to give you ownership of my own `openstack-compute`, that wouldn't work because it hits this constraint. User would have to rename an image.

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+***
+
+## Tag
+
+1. Restore points of an exisitng provisioned image created by the user in time.
+2. Performance of Tag creation will depend on the data stores' capabilities.
+3.  All tag must be deleted before de-provisioning. 
+
+* id : Id of the tag (primary key)
+* name : Name of the tag
```
knikolla: I don't really think you need ID here. A unique name would be enough, and make the pair `(name, nodeId)` unique.

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+
+***
+
+## External Service Table (Authention/DataStore/ProvisionEngine) 
+
+1. Authentication: A service used to authenticate the legitimacy of a user request.
+2. DataStore: A filesystem service where images are stored.
+3. ProvisionEngine: A service used to (de)-provision a node.
+
+* The three above tables have `<id>` (primary key), `<name>` (unique key), `<type>`, `<url>` (unique key), `<ip>`, and `<port>`
```
knikolla: I **strongly** disagree with this abstraction. If you think you can make an authentication driver work with that much information, we need to talk. Also why have `ip, port` and also `url`. I would just keep `url`, since that will include also the protocol of communication, which you are missing here.

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+
+**NOTE:** `(ip, port) pair is unique key.`
+
+***
+
+## Vlan
```
knikolla: A driver which supports multiple technologies will not work if you settle on only VLANs.

docs/redesign-specs/REST/Api.md
```diff
@@ -0,0 +1,1522 @@
...
+                "someOptionalField (Optional): {
+                        "more-fields": 12352356,
+                        ...
+                }
+    }
+
+**NOTE**: After the example, an explanation for each parameter must be provided.
+
+The response header will contain the HTTP status code:
+* 200 for success
```
knikolla: In the cases of success but no body returned in the response, please use 204.

docs/redesign-specs/REST/Api.md
```diff
@@ -0,0 +1,1522 @@
...
+* `ip`: IP address of the authentication server
+* `port`: Port of the authentication server
+* `url`: URL to connect to the authentication server
+
+## DataStore
+### create
+
+Register a data store.
+
+`POST /dataStores`
```
knikolla: Please remove the camelCasing. `datastores`

docs/redesign-specs/REST/Api.md
```diff
@@ -0,0 +1,1522 @@
...
+#### Response Body Example (Failure):
+
+    {
+        "errorMessage": "This is the reason for the encountered error"
+    }
+
+**NOTE**: `Response Body Example (Failure)` is not required for every API call
+
+#### Authentication (Optional):
+* If Authentication is required by REST API call then the Credentials/Token should be passed using `HTTP Basic Auth`.
```
knikolla: Different authentication methods have different ways to pass credentials. Please, please, do not do this. Use a middleware (like keystonemiddleware) or apache module for this.

docs/redesign-specs/DB_Schema/database.md
apoorvemohan: BMI assumes that the underlying network isolation layer (e.g. HIL) is layer-2 based.

okrieg: We should be supporting IB next year as well, in which case IP address doesn't work ...  Also eventually omnipath...  Agree with L2

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+
+## User 
+
+1. An entity using BMI. 
+2. Users will be registered to BMI by an admin via the API after BMI has been deployed. 
+
+* id : Id of the user (primary key)
+* name : Name of the user (unique key) 
+* type : Type of user (ordinary user or admin)
+* quota : Quantity of datastore disk-space allocated to a user (GB)
```
okrieg: Is this for the disk space?  I am confused by figure, where the datastore/storage provider is coupled with the database.  I think that storage provider should be closely associated with the iscsi service.  You really want a single driver for communicating with storage to clone volumes... and to set up ISCSI end points for those volumes.  For quota, is this the absolute size of the volumes, or is this the new storage incurred due to the clone; in many case the latter will be orders of magnitude less

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+* status : Current status of user (e.g. active, inactive, disabled, etc.)
```
okrieg: What does this mean?  How can you tell if someone is active?  If you have any machines currently backed by BMI then it is active, so user stays active for days/weeks/months?  What is purpose of this?

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+* autheticationId : Id of the authentication service using which a user will be verified in a multi-tenant setup. A user will specify a token (obtained from the authentication service) while invoking the REST API.
```
okrieg: Don't know enough, but that doesn't seem right to me...

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Image/Snapshot
+
+1. A virtual disk (a.k.a. golden image) whose clone (linked clone) is used to provision a node. 
+2. Users can upload images. 
+3. The user uploading the image will be the owner of the uploaded image.
+4. Snapshot is a special type of image that is created by deep copying the existing state of a provisioned node's image. 
+5. Image/Snapshot owner can be different from the owner of an ACL. 
+6. Ownership of the image/snapshot can be transfered to another user.
+7. Image/Snapshot can be shared to a set of users.
```
okrieg: if shared, and used, can the "owner" delete it?  If public, who owns it?

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+## ProvisionedNode
+
+1. Node running an operating system from the associated clone of a parent image (aka golden image). 
+2. A user can provision multiple nodes.
+3. A clone (shallow or deep copy depending on the underlying data store's capabilities) of the parent image is created and used to provision a node.
+4. Nodes are provisioned using a provisioning engine over a network owned by the user.
+5. BMI creates an interface (physical or virtual) on the provisioning engine for the users' provisioning network.
+
+* id : Id of the provisioned node (primary key)
+* macAddress : Mac address of the provisioned node (unique key)
```
okrieg: I would rather use terms like networkAddress, and say, e.g., macAddress, some day we will move away from ethernet

docs/redesign-specs/DB_Schema/database.md
```diff
@@ -0,0 +1,199 @@
...
+* parentImageId : Id of the golden image.
+* cloneName :  Name of the cloned image (unique key)
+* nic : NIC of the provisioned node 
+* ownerUserId : Id of the user who owns the provisioned node 
+* provisionEngineVlanId: Id of the provisionEngineVlan. This column represents the `(provisioning engine, vlan interface)` pair that is used to provision the given node.
```
okrieg: can we call network, and say, for example Vlan... some day move away from Vlans...

docs/redesign-specs/redesignrequirements.md
```diff
@@ -0,0 +1,29 @@
...
+
+* Remove tight coupling between CLI and BMI Components (Einstein and Database).
+
+* Enable users to replace any external components as per his requirements and resource availability (driver based model).
+  - Support for hardware iSCSI servers.
+  - Support for File/Sector based diskless provisioning.
+  - Support for different authentication services (e.g. LDAP, Kerberos, etc.).
+  - Support for different data stores (e.g. Lustre, etc).
+  - Support for different network isolators.
+
```
okrieg: I don't see the mechanisms you are using... do we have a plugin architecture?  How configure...

Issue #162, opened 2018-03-26
Get rid of calls to attach and remove networks to hil (in this version of BMI).


naved001:
This came up after running HIL/BMI in kumo for a while. 

When user give up their nodes (or we take it forcibly from them), we don't want to deprovision the nodes because that will delete their image, so we just disconnect all the networks and remove it from their project (and with a script, delete the iPXE file).

 Now if a user wants to deprovision a node, they must add that node back to their project, connect the hil_network and only then they can call the `bmi dpro` command.

Izhmash:
I'll work on this

PR #165, opened 2018-03-29
removed attach and detach network calls

Izhmash:
Addresses #162.


naved001:
Discussed this with @apoorvemohan 
we should remove the network-name argument from the CLI and from the rest call.
`bmi pro node image nic` and `bmi dpro node image nic`

PR #167, opened 2018-04-11
1nodeceph

radonm:


naved001:
I am gonna move this to some repo to keep it, will then close it here

naved001:
We moves this here https://github.com/CCI-MOC/admin-tools/tree/master/Ceph

PR #111, opened 2017-04-25
Inconsistent state

chemistry-sourabh:
@CCI-MOC/bmi-breakers initial sneak peek into the inconsistent state changes. Provision is done and also wrote some test cases, review them and suggest some more cases (Plan to write 2 more).

Deprovision in theory is done need to do it in a full proof manner and write test cases.

PR #170, opened 2018-06-22
Catch a bunch of errors everywhere:

naved001:
This PR is pretty much copy-pasta from @chemistry-sourabh's #111 PR
since dealing with the merge conflicts was nasty and it made some changes
which I now feel are not necessary.

This has some fixes for database issues, ceph image issues, and tgt targets. And it's weird that we have to catch integrity errors, the code should be such that it's never even raised (I'll look fix that within this PR or maybe a different one).

I will add some integration tests too.

@apoorvemohan @mihirborkar @izhmash

### DIFF
--- a/ims/einstein/ceph.py
+++ b/ims/einstein/ceph.py
@@ -116,7 +116,11 @@ class RBD:
                 img_name = parent_snap_name
             raise file_system_exceptions.ImageNotFoundException(img_name)
         except rbd.ImageExists:
-            raise file_system_exceptions.ImageExistsException(clone_img_name)
+            logger.info("Clone with name %s exists" % clone_img_name)
+            actual_parent = self.get_parent_info(clone_img_name)[1]
+            if actual_parent != parent_img_name:
+                raise file_system_exceptions.ImageExistsException(
+                    clone_img_name)
         # No Clue when will this be raised so not testing
         except rbd.FunctionNotSupported:
             raise file_system_exceptions.FunctionNotSupportedException()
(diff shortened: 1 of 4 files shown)


ims/einstein/iscsi/tgt.py
```diff
@@ -125,14 +125,19 @@ def remove_target(self, target_name):
         try:
             targets = self.list_targets()
             if target_name in targets:
-                os.remove(os.path.join(self.TGT_ISCSI_CONFIG,
-                                       target_name + ".conf"))
                 command = "tgt-admin -f --delete {0}".format(target_name)
                 output = shell.call(command, sudo=True)
                 logger.debug("Output = %s", output)
+                os.remove(os.path.join(self.TGT_ISCSI_CONFIG,
```
ianballou, line 131: Why did you have to move this command?

ianballou (approved):
I think this looks cool

naved001:
Thanks for the review, but I think this will require more work once #174 is merged since that restructures a lot of things.

ims/einstein/ceph.py
```diff
@@ -145,15 +149,15 @@ def list_children(self, img_id, parent_snap):
...
         except rbd.ImageNotFound:
-            raise file_system_exceptions.ImageNotFoundException(img_id)
+            logger.info("%s image is not found" % img_id)
         # Don't know how to raise this
         except rbd.ImageBusy:
             raise file_system_exceptions.ImageBusyException(img_id)
         # Forgot to test this
         except rbd.ImageHasSnapshots:
             raise file_system_exceptions.ImageHasSnapshotException(img_id)
+        return true
```
jeremyfreudberg, line 160: welll, that's something

ims/einstein/iscsi/tgt.py
```diff
@@ -125,14 +125,19 @@ def remove_target(self, target_name):
...
+                                       target_name + ".conf"))
             else:
-                raise iscsi_exceptions.TargetDoesntExistException()
-        except (IOError, OSError) as e:
+                logger.info("%s target doesnt exist" % target_name)
+        except OSError as e:
+            if "[Errno 2] No such file or directory" in str(e):
```
jeremyfreudberg, line 136: you should probably compare the value of `e.errno` to `2`, or for bonus points, compare it to `os.errno.ENOENT`

jeremyfreudberg (changes requested):

ims/einstein/ceph.py
naved001: oops

ianballou:
How's this code looking now with the recent changes? Is everything still relevant?

PR #172, opened 2018-06-29
change the name of default snapshot

### LINKED ISSUE #125: Change the default snapshot name from "snapshot" to something else.
This snapshot already exists causes a lot of confusion to lot of users. If ceph is completely untouched it should still work, but sadly that may not be the case always. So we should probably have the default name changed to some UUID,

naved001:
Fixes #125

### DIFF
--- a/ims/common/constants.py
+++ b/ims/common/constants.py
@@ -120,7 +120,6 @@ IET_LUN_STARTING = "Lun"
 DNSMASQ_LEASES_LOC = '/var/lib/misc/dnsmasq.leases'
 
 HIL_CALL_TIMEOUT = 10
-DEFAULT_SNAPSHOT_NAME = "bmi-created-snapshot"
 
 BMI_ADMIN_PROJECT = "bmi_infra"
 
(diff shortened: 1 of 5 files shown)


ims/common/constants.py
```diff
@@ -120,7 +120,7 @@
 DNSMASQ_LEASES_LOC = '/var/lib/misc/dnsmasq.leases'
 
 HIL_CALL_TIMEOUT = 10
-DEFAULT_SNAPSHOT_NAME = "snapshot"
+DEFAULT_SNAPSHOT_NAME = "bmi-created-snapshot"
```
apoorvemohan: I think this should be the UUID specified in the bmiconfig.cfg file instead of hardcoding it. What say?

naved001: does it have to be a UUID, or just any string in the config file.

apoorvemohan: i think either would work for now

naved001:
Made the change you suggested, but one of the tests is now reading from a configuration file that I don't know about. *sigh*. And that's why it's failing the tests.

ianballou (commented):
This looks fine to me, I'll +1 when tests pass.

apoorvemohan:
LGTM

apoorvemohan (approved):

naved001:
now I need to dig deep in the CI.

ianballou (approved):

naved001:
closing this in favor of #187

PR #166, opened 2018-04-04
Remove network calls

naved001:
Fixes #162 
Also, fixes the comment issue described in #128 but not any other inconsistency we may have in the CLI.

PR #174, opened 2018-07-05
WIP: Split the API

naved001:
In this PR I split the APIs so that foreman can talk to BMI. This is built on top of #166 

* Another thing this does is that, now it decouples the actual iscsi disk from the physical host. The workflow will now be something like this:
1. User creates a new disk from a golden image. (ceph clone, iscsi target add, db entry)
2. They provision a node to boot form the newly created disk image (macaddress file, ipxe file).
3. They can re-provision the node to point to a different disk image.
4. They can delete a disk_image that they no longer use.

So, foreman can now just call BMI to create disk_image and handle the other bits on it's own.

Notes:
1.  There's no de-provision command right now.
2. This breaks the APIs, so if anything was using the old BMI APIs that will have to be updated to use the new stuff. 

This is a work in progress, I still need to update everything to work with the notion of disk_images now. Opening this up so I can gather feedback on the changes before I go ahead and modify everything else. This is going to be a big one.

Things I still need to update:

- [ ] Update the rest of the methods.
- [ ] Update the REST APIs
- [ ] Update/add tests.
- [ ] Document the new APIs.
- [ ] (Easy) add delete_disk CLI call
- [ ] We should have a way to get an iscsi target without creating a disk.
- [ ] DB no longer can search for images by project, probably something to do with the schema changing here

@ianballou I updated the checklist here with the items you suggested.

cc// @Izhmash @apoorvemohan

### DIFF
--- a/ims/einstein/operations.py
+++ b/ims/einstein/operations.py
@@ -211,105 +211,108 @@ class BMI:
         self.db.close()
 
     # Provisions from HIL and Boots the given node with given image
-    @log
-    def provision(self, node_name, img_name, nic):
-        try:
-            mac_addr = "01-" + self.hil.get_node_mac_addr(node_name). \
-                replace(":", "-")
 
+    @log
+    def create_disk(self, disk_name, img_name):
+        """
+        Creates a new image named <disk_name> from <img_name>,
+        and then creates an iscsi endpoint for <disk_name>.
+        Returns the disk image name (which also happens to be the iscsi target
+        name)
+
+        Note: The name of the iscsi target and the name of disk in ceph is same
+        """
+
+        # Database Operations
+        try:
+            # Find the image id by using the image name, and then create a new
+            # entry for the new disk image.
             parent_id = self.db.image.fetch_id_with_name_from_project(img_name,
                                                                       self.proj
                                                                       )
-            self.db.image.insert(node_name, self.pid, parent_id)
-            clone_ceph_name = self.__get_ceph_image_name(node_name)
+            self.db.image.insert(disk_name, self.pid, parent_id)
+
+            # This generates the name that BMI *must* have used when it created
+            # the copy in ceph of img_name
             ceph_img_name = self.__get_ceph_image_name(img_name)
+
+            # Prepare the ceph image name to be used.
+            clone_ceph_name = self.__get_ceph_image_name(disk_name)
+        except DBException as e:
+            logger.exception('')
+            return self.__return_error(e)
+
+        # Storage Operations
+        try:
             self.fs.clone(ceph_img_name, constants.DEFAULT_SNAPSHOT_NAME,
                           clone_ceph_name)
-            ceph_config = self.cfg.fs
-            logger.debug("Contents of ceph_config = %s", str(ceph_config))
+        except FileSystemException as e:
+            logger.exception('')
+            self.db.image.delete_with_name_from_project(node_name, self.proj)
+
+        # iSCSI Operations
+        try:
             self.iscsi.add_target(clone_ceph_name)
-            logger.info("The create command was executed successfully")
-            self.__register(node_name, img_name, clone_ceph_name, mac_addr)
+        except ISCSIException as e:
+            logger.exception('')
+            self.fs.remove(clone_ceph_name)
+            self.db.image.delete_with_name_from_project(node_name, self.proj)
+
+        logger.info("The create_disk command was executed successfully")
+
+        # This will be changed to return a list of all targets, with each
+        # target including the endpoint, port and target name.
+        return self.__return_success(clone_ceph_name)
+
+    @log
+    def delete_disk(self, disk_name):
+        """
+        Deletes the disk image <disk_name>. Also deletes the iSCSI target
+        pointing to that image.
+        """
+        try:
+            # Get the ceph image name.
+            ceph_img_name = self.__get_ceph_image_name(disk_name)
+        except DBException as e:
+            logger.exception('')
+            return self.__return_error(e)
+
+        try:
+            self.iscsi.remove_target(ceph_img_name)
+        except ISCSIException as e:
+            logger.exception('')
+            return self.__return_error(e)
+
+        try:
+            ret = self.fs.remove(str(ceph_img_name).encode("utf-8"))
+        except FileSystemException as e:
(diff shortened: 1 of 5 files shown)


ianballou:
Guessing you'll want #166 to merge first

ianballou:
The work flow here definitely makes sense to me.  Random thought, would there ever be an issue if someone made a ton of disks? Is there a limit to how many iscsi targets there can be?

naved001:
Yes, we don't have quotas right now. We'll need to implement those later
(for BMI standalone).

ims/einstein/operations.py
```diff
@@ -211,119 +207,108 @@ def shutdown(self):
...
-            self.db.image.delete_with_name_from_project(node_name, self.proj)
-            time.sleep(constants.HIL_CALL_TIMEOUT)
-            self.hil.detach_node_from_project_network(node_name, network,
-                                                      nic)
-            return self.__return_error(e)
+        logger.info("The create_disk command was executed successfully")
+
+        # This will be changed to return a list of all targets, with each
+        # target including the endpoint, port and target name.
+        return self.__return_success(clone_ceph_name)
```
naved001, line 265: @apoorvemohan @Izhmash I am thinking of returning something like this
```python
>>> returnee = {"iscsi-server": "127.0.0.1","port": "3260", "target": "mytarget"}, {"iscsi-server": "192.168.3.1", "port": "3261", "target": "mytarget"}
>>> print returnee
({'target': 'mytarget', 'iscsi-server': '127.0.0.1', 'port': '3260'}, {'target': 'mytarget', 'iscsi-server': '192.168.3.1', 'port': '3261'})
>>> json.dumps(returnee)
'[{"target": "mytarget", "iscsi-server": "127.0.0.1", "port": "3260"}, {"target": "mytarget", "iscsi-server": "192.168.3.1", "port": "3261"}]'
```

ianballou: Unless there would be a case where the target dictionaries need to be named, I think this works well.

naved001:
Looks like there must be a deprovision call to delete either the ipxe file or the mac address file or both; otherwise someone else who gets my machine can just connect the network and boot off of my disk.

naved001:
Re: deprovision call

It probably doesn't matter in the foreman case, but it will matter when someone wants to use BMI standalone.

Like I said, we can delete the ipxe file or the boot configuration file (named after the mac address) or both. I wanted to run it by you guys:
1. To delete the boot configuration file, I'll need the mac-address. We could have the user provide the mac-address or provide the name of the nic. In any case, we would need to go to HIL to make sure that the nic/node with that mac-address belongs to the user.
2. We could also just simply delete the ipxe file which only requires the disk-name and bmi can verify from it's own database if the user is authorized to remove it.
3. We could do both for additional safety.

@apoorvemohan @Izhmash @mihirborkar 
what do you guys think?

ianballou:
@naved001 I think we should get rid of the ipxe file and the boot config file.  I feel like I'd get pretty confused if I deprovisioned my server but it's still running pxe commands from somewhere.  I suppose we could just take the nic name and let HIL deal with it since that's what we do in other places.

ianballou:
I gave this a test and for whatever reason the image name argument isn't making it to the create_disk function.  I'm investigating further.

ianballou:
Looks like we're missing a `constants.DISK_NAME_PARAMETER` in the create_disk API call.

ims/picasso/rest.py
```diff
@@ -110,3 +103,15 @@ def list_snapshots():
            [constants.IMAGE_NAME_PARAMETER])
 def remove_image():
     pass
+
+
+@rest_call("/create_disk", "PUT", constants.CREATE_DISK_COMMAND,
+           [constants.DISK_NAME_PARAMETER])
```
ianballou: Looks like we're missing a constants.DISK_NAME_PARAMETER in the create_disk API call.

ianballou (commented):
We also need to remote the provisioning network from the list projects call

ianballou:
Another thing: we need to add the command names to rpc

ims/einstein/operations.py
```diff
@@ -211,119 +207,108 @@ def shutdown(self):
...
-
-        except RegistrationFailedException as e:
-            # Message is being handled by custom formatter
-            # TODO: add a deployment and a unit test for this case.
+        except FileSystemException as e:
             logger.exception('')
-            clone_ceph_name = self.__get_ceph_image_name(node_name)
-            self.iscsi.remove_target(clone_ceph_name)
-            self.fs.remove(clone_ceph_name)
             self.db.image.delete_with_name_from_project(node_name, self.proj)
```
ianballou, line 251: `node_name` is undefined

ims/einstein/operations.py
```diff
@@ -211,119 +207,108 @@ def shutdown(self):
...
 
+        # iSCSI Operations
+        try:
+            self.iscsi.add_target(clone_ceph_name)
         except ISCSIException as e:
-            # Message is being handled by custom formatter
             logger.exception('')
-            clone_ceph_name = self.__get_ceph_image_name(node_name)
             self.fs.remove(clone_ceph_name)
             self.db.image.delete_with_name_from_project(node_name, self.proj)
```
ianballou, line 259: `node_name` undefined again

naved001:
thanks for the review @ianballou! I'll (hopefully) be back in office tomorrow and then we can fix these all.

ianballou:
@naved001 no prob! So the main things I found that need fixing:
1.  DB no longer can search for images by project, probably something to do with the schema changing here.
2. We should have a way to get an iscsi target without creating a disk.  
3. (Easy) add delete_disk CLI call

ims/einstein/operations.py
```diff
@@ -89,10 +89,6 @@ def __check_admin(self):
     def __get_ceph_image_name(self, name):
         img_id = self.db.image.fetch_id_with_name_from_project(name,
                                                                self.proj)
```
naved001: I messed up rebasing somewhere which is this deleted.

naved001:
rebased this with all commits squashed because I messed something up. Should be clean now though.

ianballou:
This worked pretty well in my Foreman integration POC.  Only thing was I needed a way to grab the target without cloning the disk again.  Some kinda check in `create_disk` if the disk was already cloned could just return the target.

PR #178, opened 2018-08-16
remove a column

naved001:
because we don't have networks anymore. @apoorvemohan this is what you needed.


ianballou (approved):

jeremyfreudberg (approved):
this also appears entirely reasonable

naved001:
Cool, thanks. Just rebasing this and will merge it then.

PR #179, opened 2018-08-20
Use massopencloud docker registry rathen than bmis because:

naved001:
* we dont have access to bmis docker registry
* we probably want to keep things on massopencloud repos/registries.

Will come in handy when we need to update the docker images. @apoorvemohan @ianballou I can add your accounts to massopencloud docker repo/registry (or whatever they call it)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,10 +19,10 @@ before_script:
   - sleep 30
 
   # Run HIL Container
-  - sudo docker run -d --name hil bmis/hil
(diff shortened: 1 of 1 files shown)


ianballou (approved):
Looks good to me.  Cool we'll be in control of those containers now

jeremyfreudberg (approved):
this is entirely reasonable

naved001:
Thanks, merging.

PR #180, opened 2018-08-21
Use ceph-hammer container from massopencloud docker registry

naved001:
because it was removed from the the ceph registry for reasons I don't know.
So I uploaded the image to our registry from my PC.

We should probably use a newer version of ceph (either make our own image, or
get it from somewhere) later.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_script:
   - find . -name \*.py -exec python -m py_compile {} +
 
   # Run Ceph Container and Wait for some time so that ceph runs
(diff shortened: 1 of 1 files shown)


naved001:
@ianballou @jeremyfreudberg

ianballou (approved):
Looks fine to me

jeremyfreudberg (approved):
It would still be good to find out why ceph/demo was removed from Dockerhub. But this will work, to unwedge the CI and allow further dev work to happen in the meantime.

PR #183, opened 2018-09-06
Concatenate base_url and api in hil client

naved001:
* instead of using urllib's urljoin because it doesn't pick up the hil api
version from the config file.

https://stackoverflow.com/questions/10893374/python-confusions-with-urljoin

### DIFF
--- a/ci/bmiconfig.cfg.ci
+++ b/ci/bmiconfig.cfg.ci
@@ -22,7 +22,7 @@ fs = <driver to load>
 
 # This section is for network isolator (HIL) related config
 [net_isolator]
-url = http://172.17.0.3:7000/v0/
+url = http://172.17.0.3:7000
 
 # This section is for iscsi related config
 [iscsi]
--- a/ci/bmiconfig.cfg.ci
+++ b/ci/bmiconfig.cfg.ci
@@ -1,0 +1,70 @@
+# this section is for basic bmi settings
+[bmi]
+# uid is given so that images dont clash in ceph pool
+uid = 1
+service = False
+
+# this section is for db settings
+[db]
+path = /var/lib/bmi/bmi.db
+
+# This section is filesystem related config
+[fs]
+id = admin
+pool = rbd
+conf_file = /etc/ceph/ceph.conf
+keyring = /etc/ceph/ceph.client.admin.keyring
+
+[driver]
+net_isolator = <driver to load>
+iscsi = <driver to load>
+fs = <driver to load>
+
+# This section is for network isolator (HIL) related config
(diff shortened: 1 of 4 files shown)


naved001:
I don't know why these tests are failing. :/

ianballou:
Steps to update HIL for trying to fix the CI:

- [x] Use newer hil-v03 container
- [ ] Update HIL endpoint to http://ip:port/v0/
- [ ] Add Dockerfile and related scripts

ianballou:
Note: the first test that fails is a `PUT` to `create_snapshot/`.  Since there are cascading failures due to the RPC servers not quitting, this is the only test we can count on being real until it is fixed.

naved001:
@ianballou 

I added a variable to append the HIL API version in the BMI HIL client (rather than getting that from the config file). I figured out why urljoin was replacing the `v0`  in my `base_url`, that was because `api` that was passed to `urljoin` was an absolute path. So now it can talk to the new container.

Also, I downloaded the bmi container to get the config file out of it, and I am now manually copying it. That is kind of unrelated to this PR, but since I had already made that change I am gonna let that stay.

ianballou:
Oh awesome! I was about to try out the new container, glad it's working

ci/bmiconfig.cfg.ci
```diff
@@ -0,0 +1,70 @@
...
+keyring = /etc/ceph/ceph.client.admin.keyring
+
+[driver]
+net_isolator = <driver to load>
+iscsi = <driver to load>
+fs = <driver to load>
+
+# This section is for network isolator (HIL) related config
+[net_isolator]
+url = http://172.17.0.3:7000
```
ianballou, line 25: Do we know if this IP is guaranteed? I think we can assign this as a static ip when we run the container

ianballou (commented):
I just want to check out the container IP thing, but otherwise let's get this merged soon since the CI finally works again!

ci/bmiconfig.cfg.ci
naved001: In our CI we start the the ceph container first which gets the ip of 172.17.0.2 (see MON_IP in .travis.yml when launching the ceph container). And then we launch the hil container which gets the next ip which is 172.17.0.3. Right now these ips are not statically assinged (since if you launch the hil container first it gets 172.17.0.2). We can change to static assignment in a different PR.

ianballou: ah ok, I was wondering if the ips were sequential like that.  We can revisit the static bit later

ianballou (approved):

PR #184, opened 2018-09-06
Get right mac address from HIL

### LINKED ISSUE #169: BMI always gets the mac address of the first nic, irrespective of what you specify on the command line
https://github.com/CCI-MOC/ims/blob/master/ims/einstein/hil.py#L124

This should have been searching the right nic by using the label entered at the CLI.

I have already opened another issue to update the HIL code here to use the client library rather than the API directly. So doing that cleanly will also fix this issue.

Or we might just get rid of the nic label parameter and ask the user to enter the mac address directly (I think this is somewhere in the design docs).

Thanks @mihirborkar for finding this out.

naved001:
Fixes #169, this has the commit from #183

### DIFF
--- a/ims/einstein/hil.py
+++ b/ims/einstein/hil.py
@@ -105,11 +105,14 @@ class HIL:
         return self.__call_rest_api_with_body(api=api, body=body)
 
     @log
-    def get_node_mac_addr(self, node):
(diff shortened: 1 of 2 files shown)


ianballou (approved):
Alrighty, this is good

PR #187, opened 2018-09-26
Change default snapshot name

naved001:
I figured out what the problem was with #172 but I can't push to that anymore (since I deleted my fork). I already have 2 approvals from there, but it might be a good idea to look at it once again.
cc: @ianballou @apoorvemohan


naved001:
I rebased this one on top of #183 which allows me to modify the bmi config file.

naved001:
rebased.

ianballou (approved):
alrighty, looks fine to me

PR #190, opened 2018-10-03
Reorganize CI related files

naved001:


ianballou (approved):

ianballou:
Need a rebase here then we're all set

PR #191, opened 2018-10-11
Added hil ci container docker sources

ianballou:
Now it'll be easy to know just what's in the HIL CI container.  To test, `cd` to the `hil_container` directory and run `sudo docker build -t hil-v03 .`

### DIFF
--- a/ci/docker_sources/hil_container/Dockerfile
+++ b/ci/docker_sources/hil_container/Dockerfile
@@ -1,0 +1,73 @@
+FROM centos
+
+RUN yum install -y epel-release
+
+RUN yum -y update && yum install -y bridge-utils  gcc  httpd  ipmitool libvirt libxml2-devel  libxslt-devel  mod_wsgi net-tools python-pip python-psycopg2 python-virtinst qemu-kvm telnet vconfig virt-install
+
+RUN yum install -y git
+
+RUN yum install -y python-devel
+
+RUN pip install dumb-init
+
+RUN useradd -ms /bin/bash hil
+
+RUN passwd -d hil
+
+RUN passwd -d root
+
+RUN usermod -aG wheel hil
+
+WORKDIR /home/hil
+
+RUN git clone https://github.com/CCI-MOC/hil.git
+
+WORKDIR /home/hil/hil
+
+RUN git reset --hard f88d2fc92066c1e83e428ebeeca351971f306053
+
+RUN pip install --upgrade pip setuptools
+
+RUN python setup.py install
+
+COPY hil.cfg /etc/hil.cfg
+
+COPY hil.cfg /home/hil
+
+COPY hil.cfg /home/hil/hil
+
+USER hil
+
+RUN hil-admin db create
+
+USER root
+
+RUN chmod 666 /home/hil/hil.db
+
+RUN chown hil:hil /home/hil/hil.db
+
+COPY setupdb.sh setupdb.sh
+
+RUN chmod a+x setupdb.sh
+
+USER hil
+
+ENV HIL_ENDPOINT=http://127.0.0.1:7000
+
+RUN ./setupdb.sh
+
+USER root
+
+RUN cp hil.wsgi wsgi.py
+
+RUN pip install gunicorn
+
+COPY runhil.sh runhil.sh
+
+RUN chmod a+x runhil.sh
+
+EXPOSE 7000
+
+user hil
+
+CMD dumb-init /home/hil/hil/runhil.sh
--- a/ci/docker_sources/hil_container/Dockerfile
+++ b/ci/docker_sources/hil_container/Dockerfile
@@ -20,19 +20,15 @@ RUN usermod -aG wheel hil
 
 WORKDIR /home/hil
 
-RUN git clone https://github.com/CCI-MOC/hil.git
+RUN git clone https://github.com/cci-moc/hil -b v0.3
 
 WORKDIR /home/hil/hil
-
-RUN git reset --hard f88d2fc92066c1e83e428ebeeca351971f306053
 
 RUN pip install --upgrade pip setuptools
 
 RUN python setup.py install
 
 COPY hil.cfg /etc/hil.cfg
-
-COPY hil.cfg /home/hil
 
 COPY hil.cfg /home/hil/hil
 
@@ -54,7 +50,7 @@ USER hil
 
 ENV HIL_ENDPOINT=http://127.0.0.1:7000
 
-RUN ./setupdb.sh
+RUN source ./setupdb.sh
 
 USER root
 
--- a/ci/docker_sources/hil_container/hil.cfg
+++ b/ci/docker_sources/hil_container/hil.cfg
@@ -1,0 +1,36 @@
+[general]
+log_level = debug
+
+[auth]
+require_authentication = False
+
+[headnode]
+trunk_nic = eth0
+base_imgs = img1, img2, img3, img4
+libvirt_endpoint = qemu:///system
+
+[client]
+endpoint = http://127.0.0.1:7000
+
+[database]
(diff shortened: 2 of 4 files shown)


ci/docker_sources/hil_container/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+
+RUN usermod -aG wheel hil
+
+WORKDIR /home/hil
+
+RUN git clone https://github.com/CCI-MOC/hil.git
+
+WORKDIR /home/hil/hil
+
+RUN git reset --hard f88d2fc92066c1e83e428ebeeca351971f306053
```
naved001: We can directly clone the tagged version instead of doing a `git reset --hard`
`git clone https://github.com/cci-moc/hil -b v0.3`

ci/docker_sources/hil_container/Dockerfile
```diff
@@ -0,0 +1,73 @@
...
+
+RUN pip install --upgrade pip setuptools
+
+RUN python setup.py install
+
+COPY hil.cfg /etc/hil.cfg
+
+COPY hil.cfg /home/hil
+
+COPY hil.cfg /home/hil/hil
```
naved001, line 33: why are we copying it to three places?

ci/docker_sources/hil_container/hil.cfg
```diff
@@ -0,0 +1,36 @@
...
+hil.ext.switches.mock =
+hil.ext.obm.mock =
+
+#hil.ext.network_allocators.null =
+
+# Depending on what you're trying to do, you may want to use the vlan_pool
+# network allocator instead of the null allocator. To do this, comment out the
+# null allocator extension above, and uncomment the following:
+#
+hil.ext.auth.null =
```
naved001: BMI can work with database authentication, in fact there are some tests in BMI that check for that.

ci/docker_sources/hil_container/runhil.sh
```diff
@@ -0,0 +1,4 @@
+nohup gunicorn -w 5 -b 0.0.0.0:7000 wsgi:application >/dev/null 2>&1 &
+sleep 1
+nohup hil-admin serve_networks >/dev/null 2>&1 &
```
naved001: the command is `hil-admin serve-networks` we have dashes everywhere instead of an underscore.

ci/docker_sources/hil_container/setupdb.sh
```diff
@@ -0,0 +1,27 @@
+hil-admin run-dev-server -p 7000 &
```
naved001, line 5: kind of confused, why are you running it here again when you already run gunicorn on port 7000 in `runhi.sh`

ci/docker_sources/hil_container/setupdb.sh
```diff
@@ -0,0 +1,27 @@
...
+
+sleep 5
+
+hil-admin serve-networks &
```
naved001, line 9: same thing with this guy.

naved001 (changes requested):
Just a few questions, otherwise it looks fine.

ci/docker_sources/hil_container/runhil.sh
ianballou: huh, makes me wonder why this even works then

ci/docker_sources/hil_container/setupdb.sh
ianballou: This part is just to turn on hil so the database could get setup.  I've tried to keep it close as possible to the old hil container setup too

naved001: wouldn't this complain that the something is already running on port 7000?

ci/docker_sources/hil_container/runhil.sh
ianballou: I think for some reason upon exiting the bash script the two servers get killed.  I tried to do a `kill %1 %2` but it said there was nothing running

naved001: I don't think it works. 
```
[centos@hil etc]$ hil-admin serve_networks > /dev/null 2>&1 &
[1] 27435
[centos@hil etc]$ 
[centos@hil etc]$ 
[centos@hil etc]$ 
[1]+  Exit 2                  hil-admin serve_networks > /dev/null 2>&1

```
Since we push the output to /dev/null we don't see anything on screen. and it exits in a few seconds.

ianballou: I'll take out the gunicorn bit and see if the server is still running

ci/docker_sources/hil_container/setupdb.sh
ianballou: well this bit gets run before the gunicorn server, so it should be the gunicorn one that's complaining. Since we throw out the output though I think we just don't see it

ci/docker_sources/hil_container/runhil.sh
ianballou: There must be some magic going on here because if I take out the gunicorn server I can't curl to hil at all

ci/docker_sources/hil_container/hil.cfg
ianballou: wouldn't that mean we'll need to do a hil-admin call to add the first administrative user? bmi_infra I'm guessing
edit: found it

ci/docker_sources/hil_container/runhil.sh
naved001: Sorry if I wasn't clear. The gunicorn server is the api sever and that should stay. Without it, you won't be able to curl to hil at all (unless you are running the dev server using the hil-admin command).

The serve-networks command is the process that does the networking operations, if it's not running and you queue a networking operation, that operation wouldn't go through. You would still be able to curl to hil and do other things. so the bit that should get removed is `hil-admin serve_networks`... since that is a command which doesn't exist. And you are running the right command in `setupdb.sh` already.

ci/docker_sources/hil_container/setupdb.sh
naved001: okay. I am fine with either running the flasks' built-in dev server (using the hil-admin command) or the gunicorn server. But we shouldn't run both because it leads to confusion and only one will fail anyway.

ci/docker_sources/hil_container/hil.cfg
naved001: you are right, we would have to bootstrap the database somewhere.

ci/docker_sources/hil_container/runhil.sh
ianballou: Right I get that, but I believe the two server commands I run in that setupdb script get killed somewhere along the way, which is why it's fine to first run them there, setup the db, then run the servers again at the end.  Since the API is not reachable even within the container when taking out the gunicorn command, the original run-dev-server call along with serve-networks must be getting stopped. Also I fixed the serve-networks command to be correct in the latest commit

ci/docker_sources/hil_container/Dockerfile
ianballou: fixed this, only need it in two places

ci/docker_sources/hil_container/setupdb.sh
ianballou: I think I address this in the thread above, but these commands in setupdb get killed somewhere along the way before gunicorn & the second serve-networks call

ianballou: Yeah probably has to do with why runhil.sh has the `sleep infinity` in it.  I think when `setupdb.sh` exits even though there's the `&` it kills the servers.
Edit: might also be when I switch users the terminal closes

naved001 (approved):
Looks good to me

naved001:
We need to rebase this.

PR #193, opened 2018-10-29
WIP: Add multipathing

naved001:
This basically works. It configures the tgt servers correctly, and generates the right ipxe files.

Things I need to do:

- [ ] Document - everything; the location of template files, config files, etc.
- [ ] Tests - TGT driver, (might need to spawn another tgt container in testsuite).
- [x] load balance - use database to keep track of how many nodes are booting from each server.

### DIFF
--- a/ims/common/bmi_config.py
+++ b/ims/common/bmi_config.py
@@ -10,6 +10,7 @@ def parse_config(cfg):
 
     # Template section
     cfg.option(constants.TEMPLATES_SECTION, constants.TGT_TEMPLATE_FILE)
+    cfg.option(constants.TEMPLATES_SECTION, constants.IPXE_TEMPLATE_FILE)
 
     # DB Section
     cfg.option(constants.DB_SECTION, constants.DB_PATH_OPT)
--- a/ims/common/bmi_config.py
+++ b/ims/common/bmi_config.py
@@ -39,8 +39,12 @@ def parse_config(cfg):
     cfg.option(constants.DRIVER_SECTION, constants.ISCSI_DRIVER_OPT)
     cfg.option(constants.DRIVER_SECTION, constants.FS_DRIVER_OPT)
 
-    # Mandatory Sections (Typically Driver Sections)
+    # ISCSI section
     cfg.section(constants.ISCSI_SECTION)
+    cfg.option(constants.ISCSI_SECTION, constants.PRIMARY_ISCSI_HOST)
+    cfg.option(constants.ISCSI_SECTION, constants.SECONDARY_ISCSI_HOST)
+
+    # Mandatory Sections (Typically Driver Sections)
     cfg.section(constants.NET_ISOLATOR_SECTION)
     cfg.section(constants.FS_SECTION)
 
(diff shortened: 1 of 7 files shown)


naved001:
I don't have to worry about load balancing in code here I think because multipathd will do that automatically.

naved001:
Also, this pretty much works in testing with real hardware. Exceptions were raised and handled correctly.

ims/common/bmi_config.py
```diff
@@ -36,8 +40,12 @@ def parse_config(cfg):
     cfg.option(constants.DRIVER_SECTION, constants.ISCSI_DRIVER_OPT)
     cfg.option(constants.DRIVER_SECTION, constants.FS_DRIVER_OPT)
 
-    # Mandatory Sections (Typically Driver Sections)
+    # ISCSI section
     cfg.section(constants.ISCSI_SECTION)
+    cfg.option(constants.ISCSI_SECTION, constants.PRIMARY_ISCSI_HOST)
+    cfg.option(constants.ISCSI_SECTION, constants.SECONDARY_ISCSI_HOST)
```
ianballou, line 46: Do we only need a primary and secondary host?

ims/einstein/iscsi/tgt.py
```diff
@@ -101,57 +111,102 @@ def add_target(self, target_name):
...
-            raise iscsi_exceptions.TargetCreationFailed(str(e))
-        except shell_exceptions.CommandFailedException as e:
-            raise iscsi_exceptions.TargetCreationFailed(str(e))
+        targets = self.list_targets()
+        if target_name in targets:
+            raise iscsi_exceptions.TargetExistsException()
+
+        self.__generate_config_file(target_name)
+
+        command = "sudo tgt-admin --execute"
```
ianballou, line 120: Why do we need to `sudo` now when we didn't before?

ims/einstein/iscsi/tgt.py
```diff
@@ -29,70 +41,68 @@ def start_server(self):
...
         try:
-            status = shell.get_service_status('tgtd')
-            if status not in ['Running', 'Dead']:
-                return "Error"
-        except shell_exceptions.CommandFailedException:
-            raise iscsi_exceptions.ShowStatusFailed()
+            cmds = self.client.copy_file(
+                config_file_path, self.TGT_ISCSI_CONFIG + config_file_name)
+            if self.secondary_iscsi is not None:
+                joinall(cmds, raise_error=True)
```
ianballou, line 95: Why do we need this `joinall` here?

ianballou (commented):
Just a few questions, I really need to test this out on some hardware.

ims/common/bmi_config.py
naved001: Yep, I can't specify more than 2 paths anyway in the ipxe file.

ims/einstein/iscsi/tgt.py
naved001: because earlier the code had to be run as root...by specifying sudo I think we can make an entry in sudoers file for the bmi-user to only execute tgt-admin commands. I am open to other ideas if there's a better way

ims/einstein/iscsi/tgt.py
naved001: Per the docs here https://github.com/ParallelSSH/parallel-ssh#sftp
If I am copying to both iscsi hosts, I want to wait for both threads to finish

naved001:
tgt rbd don't work for mp https://ceph.com/planet/ceph-rbd-and-iscsi/

